PR #91, opened 2022-03-10
Fix naming to reflect sansshell

sfc-gh-jchacon:


sfc-gh-ajoyner (approved):
Please add note in the squashed commit description referencing commit 0ef6d53 for more background.

PR #93, opened 2022-03-25
Updated README.md

sfc-gh-iuion:
Update go install instructions, which changed with the go1.18. Also fix minor typo.


sfc-gh-ajoyner (approved):
Thanks, Iulia!

PR #94, opened 2022-03-29
When we create streams we may have targets not connect.

sfc-gh-jchacon:
Track those in a separate errors list of precanned errors to return on places we process replies.

i.e. InvokeOneMany just emit before the loop.

For streams we do it in Recv but make sure to only emit the errors once or else we spam a lot (each packet/target)

Add tests to make sure it acts sanely


sfc-gh-jallie (approved):
LGTM

PR #96, opened 2022-04-04
Add support for loaders being able to implement reloadable credentials.

sfc-gh-jchacon:
Do this by returning a wrapped credentials.TransportCredentials that calls a new method on the loader: CertsRefreshed.

Implementations should check when they internally reload their certificates and then make certs refreshed true in those cases.

This will then make future calls using the credentials to generate a new TransportCrendentials before pass through calls to using them.

Add unit tests and restructure a bit (we weren't always using client vs server certs when we said we were).


auth/mtls/mtls.go
```diff
@@ -55,6 +58,70 @@ type CredentialsLoader interface {
...
+
+type WrappedTransportCredentials struct {
+	creds      credentials.TransportCredentials
+	ctx        context.Context
+	loaderName string
+	loader     func(context.Context, string) (credentials.TransportCredentials, error)
+}
+
+func (w *WrappedTransportCredentials) checkRefresh() error {
+	loader, err := Loader(w.loaderName)
```
sfc-gh-jallie, line 77: do we actually need to call Loader() on each iteration? I suppose it's "correct", since something could technically register a new loader impl, but feels like we could just hang on to it after the first call

auth/mtls/mtls.go
```diff
@@ -55,6 +58,70 @@ type CredentialsLoader interface {
 	// LoadServerCertificate returns the certificate that should be presented
 	// by the server to incoming clients.
 	LoadServerCertificate(context.Context) (tls.Certificate, error)
+
+	// CertRefreshed indicates if internally any of the cert data has
+	// been refreshed and should be reloaded. This will depend on the
+	// implementation to support but allows for dynamic refresh of certificates
```
sfc-gh-jallie, line 64: I suppose either the implementation will need to do work (including maybe file I/O) in this call context, or have an internal go-routine to set an 'update' flag. I wonder if we want to make some sort of implementation recommendation here?

auth/mtls/mtls.go
```diff
@@ -55,6 +58,70 @@ type CredentialsLoader interface {
...
+	// without a server restart.
+	CertsRefreshed() bool
+}
+
+type WrappedTransportCredentials struct {
+	creds      credentials.TransportCredentials
+	ctx        context.Context
```
sfc-gh-jallie, line 71: can this field be removed?

auth/mtls/mtls.go
```diff
@@ -55,6 +58,70 @@ type CredentialsLoader interface {
...
+}
+
+// Info -- see credentials.Info
+func (w *WrappedTransportCredentials) Info() credentials.ProtocolInfo {
+	w.checkRefresh()
+	return w.creds.Info()
+}
+
+// Clone -- see credentials.Clone
+func (w *WrappedTransportCredentials) Clone() credentials.TransportCredentials {
```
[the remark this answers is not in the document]
sfc-gh-jchacon: good point. Did a clone of it

auth/mtls/mtls.go
```diff
@@ -55,6 +58,70 @@ type CredentialsLoader interface {
...
+	w.checkRefresh()
+	return w.creds.Clone()
+}
+
+// OverrideServerName -- see credentials.OverrideServerName
+func (w *WrappedTransportCredentials) OverrideServerName(s string) error {
```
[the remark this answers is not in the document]
sfc-gh-jchacon: yeah, did that. Stored it in the wrapped one and set it on each refresh (and if called directly).

auth/mtls/mtls.go
sfc-gh-jchacon: Yeah, was torn here. Technically to generate new credentials we need a context so it was either cache one or use context.Background(). I removed it and used context.Background instead which feels a little cleaner since cached ones might be expired.

Mainly because the credentials interface doesn't pass creds through each function, only some.

auth/mtls/mtls.go
sfc-gh-jchacon: It's more complicated. We don't know what kind of creds we're generating (client vs server) which is what loader is handling. The only interface we have for that has to lookup the loader each time then as we call it. I couldn't figure out another way here.

auth/mtls/mtls.go
sfc-gh-jchacon: Yeah...the credential loaders we provide just return false always so this never happens. It really is context driven i.e. I mocked up a vault PR based on this and it was exactly that where we track and updated state and clear it once it gets checked.

I'm not sure what we could recommend here without just providing something like vault as an example?

auth/mtls/mtls.go
sfc-gh-jchacon: ok, slept on it and more obvious. Fixed

sfc-gh-jallie (approved):
LGTM

PR #97, opened 2022-04-05
Add an extensions field in RPCAuthInput.

sfc-gh-jchacon:
Allows implementation defined additions to auth input for passing into OPA for evaluation.

Add tests showing this works as expected.

### DIFF
--- a/auth/opa/rpcauth/input.go
+++ b/auth/opa/rpcauth/input.go
@@ -52,7 +52,7 @@ type RPCAuthInput struct {
 	Host *HostAuthInput `json:"host"`
 
 	// Implementation specific extensions.
-	Extensions []interface{} `json:"extensions"`
+	Extensions json.RawMessage `json:"extensions"`
 }
 
 // PeerAuthInput contains policy-relevant information about an RPC peer.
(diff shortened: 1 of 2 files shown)


auth/opa/rpcauth/input.go
```diff
@@ -45,11 +45,14 @@ type RPCAuthInput struct {
...
 
-	// Information about the calling peer, if available
+	// Information about the calling peer, if available.
 	Peer *PeerAuthInput `json:"peer"`
 
 	// Information about the host serving the RPC.
 	Host *HostAuthInput `json:"host"`
+
+	// Implementation specific extensions.
+	Extensions []interface{} `json:"extensions"`
```
sfc-gh-jallie, line 55: as mentioned in chat:

I think this might be nicer as json.RawMessage, rather than []interface

reasoning:
 - it makes it clear that the value needs to be convertible to json (and sidesteps ensuring that it's cleanly convertible, and doesn't have some custom marshaller that runs for infinite time or panics)
 - it's "cheaper", in the case that this is just some static list of attributes (imagine: additional host attributes/tags/etc) which only need to be marshalled once, and then just stuffed in by a hook (MarshalJSON for RawMessage is a noop)

sfc-gh-jallie (approved):
fine as is, but see comment re: json.RawMessage vs interface with reasoning for why we might want to switch

auth/opa/rpcauth/input.go
sfc-gh-jchacon: Yep, as discussed cleaner API for end users so converted over.

PR #98, opened 2022-04-06
Client side OPA policy implementation

sfc-gh-jchacon:
grpc will guarentee you the other side of an MTLS passed authn (i.e. it's a validly signed cert) but not necessarily that it's the cert you expected to see.

Client side OPA allows validation of any cert field we desire to give authz for whether we really accept/trust the server to be who it claims to be.

Provide an example in the integration test of both passing and failing.

Plumb logging into sanssh (off by default without -v) to allow debugging


sfc-gh-jallie (approved):
LGTM

PR #99, opened 2022-04-12
Add some needed exec options:

sfc-gh-jchacon:
Run as a specific user/group
Append specific env vars to the otherwise clean environment.

Tests for everything

### DIFF
--- a/services/util/util.go
+++ b/services/util/util.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/Snowflake-Labs/sansshell/proxy/proxy"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


services/util/util.go
```diff
@@ -53,12 +54,14 @@ type CommandRun struct {
...
 
-// There's only one now but compose into a builder pattern for Options so
-// it's easy to add more (such as dropping permissions, etc).
+// A builder pattern for Options so it's easy to add various ones (such as dropping permissions, etc).
 type cmdOptions struct {
 	failOnStderr bool
 	stdoutMax    uint
 	stderrMax    uint
+	env          []string
+	uid          *uint32 // Pointers so we can distinguish if it's been set.
```
sfc-gh-jallie, line 63: since you never let the caller construct these directly, you could probably just set these to defaults before you apply options:

```go
 o := cmdOptions{
   uid: os.Geteuid(),
   gid: os.Getgid()
}
for _, opt := range options {
   opt(o)
}

// use options here, always constructing a credential
```

sfc-gh-jallie (approved):
LGTM

services/util/util.go
sfc-gh-jchacon: True. Still have to validate they differ because it's EPERM to call setuid/gid() even if they are the values you already are...

sfc-gh-jchacon: Fixed

PR #100, opened 2022-04-12
Add support to file ops so we can do things by name instead of requiring uid/gid.

sfc-gh-jchacon:
i.e. setting permissions with chown or copying files up.

This allows easier use from the CLI where one may not know the uid/gid already.

The server will lookup values when presented with names.

The options are exclusive. Can't set both uid and username together.

Flesh out some tests and integration tests to validate.


sfc-gh-jallie (approved):
LGTM

PR #101, opened 2022-04-13
Update dependency packages

sfc-gh-jchacon:


sfc-gh-jelsesiy (approved):
LGTM, do you want to bump to go1.18 as well?

sfc-gh-jchacon:
Will do that in a followup PR after I make sure this works on our internal copy (and update there)

PR #103, opened 2022-04-26
Add kill to RPCs in process module

sfc-gh-jchacon:
Allow sending a single pid a given signal (including 0).

Just return nothing or use the error if something was wrong (permissions, etc)

Add tests and integration.


sfc-gh-jallie (approved):
LGTM

PR #104, opened 2022-04-27
Fix handling of --proxy and --targets for IPv6 addresses or empty strings

l9i:
There are two problems with the current logic of appending the default port:
- Specifying an empty `--proxy` flag will result in appending the default proxy port.
- Specifying an IPv6 address in `--proxy` or `--targets` without a port will not result in appending the default port.

### DIFF
--- a/testing/integrate.sh
+++ b/testing/integrate.sh
@@ -576,7 +576,7 @@ EOF
 ${SANSSH_PROXY} ${SINGLE_TARGET} --v=1 --client-policy-file=${LOGS}/client-policy.rego healthcheck validate
 check_status $? policy check should succeed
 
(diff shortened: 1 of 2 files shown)


sfc-gh-jchacon (approved):
Nice logic in hasPort

testing/integrate.sh
```diff
@@ -574,6 +576,9 @@ EOF
 ${SANSSH_PROXY} ${SINGLE_TARGET} --v=1 --client-policy-file=${LOGS}/client-policy.rego healthcheck validate
 check_status $? policy check should succeed
 
+${SANSSH_NOPROXY_NOPORT} ${SINGLE_TARGET_NOPORT} healthcheck validate
```
sfc-gh-jchacon, line 579: This should be SANSSH_PROXY_NOPORT based on vars above?

sfc-gh-jchacon (approved):

PR #107, opened 2022-05-03
Add a jitter param to the basic server startup and plumb it into the main server

sfc-gh-jchacon:
The server will be deployed across a large swath of machines potentially.
If a large number of nodes start/restart at once we don't want thundering herd problems if they have credential loading issues (say a remote cred store).

So allow the user to plumb a jitter in to the server.Run which can use this before it exits.

Add integration test to prove it works.


sfc-gh-jchacon:
We're going to drop this as we can just handle this in a startup wrapper instead

PR #108, opened 2022-05-04
Add a version RPC and CLI arg for returning the embedded one.

sfc-gh-jchacon:
Expose in CLI/proxy-server as it's own method (similar to logging).

Integrate both methods into integration test


sfc-gh-jallie (approved):
LGTM

PR #111, opened 2022-05-06
Remove hard coded timeout on context.

sfc-gh-jchacon:
If the incoming was lower than 30s this is ignored anyways I beleive. Plus the top level always sets one with a deadline.

Sansshell module obviously cargo culted from healthcheck module.


sfc-gh-jallie (approved):
LGTM

PR #113, opened 2022-05-09
Make sure we include file support for blobs.

sfc-gh-jchacon:
In the localfile client augment cp so it can take file:// style buckets to do local copies on machines (remotely).


sfc-gh-hdhillon (approved):
lgtm

PR #115, opened 2022-05-10
Check the args left to be processed and if they are flags or help assume this is an internal command and just process it

sfc-gh-jchacon:
Otherwise we have to set proxy/target just to get help which feels weird.

Remve leftover comment while here


sfc-gh-jallie (approved):
LGTM. This also bugged me, so I'm glad to see if fixed

PR #116, opened 2022-05-10
Provide an example of how to override the flag default for a service flag

sfc-gh-jchacon:


sfc-gh-jchacon:
Let me take another pass. I might just convert everything to vars and bind in main()

PR #117, opened 2022-05-10
FDB cli implementation for sansshell

sfc-gh-jchacon:
Wrap the fdbcli command and implement each subcommand as distinct proto option. This allows setting policy on specific items without having to give either blanket access (run this set of string commands) or hard to parse policy doing string matching.


sfc-gh-ssudakovich (approved):
LGTM

PR #119, opened 2022-05-12
Proxy: remove blocking during StartStream

sfc-gh-jallie:
# Description

This PR updates the proxy to avoid blocking StartStream for unconnectable target hosts.

Previously, when processing a new StartStream requests, the proxy code would first grpc.Dial the target (which is non-blocking by default), and then create a new client stream using conn.NewStream (which _is_ blocking).

The net effect of this is that a single unreachable host could block requests for all targets up to the timeout of the client->proxy RPC

To avoid this, the updated code defers making the blocking call to NewStream until work has been moved onto the per-target go-routine.

# Testing

Added a new test which ensures that TargetStreamSet.Add does not block (which fails under the previous version of the code).

### DIFF
--- a/proxy/server/target.go
+++ b/proxy/server/target.go
@@ -18,6 +18,7 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand"
@@ -28,6 +29,7 @@ import (
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/anypb"
@@ -39,6 +41,9 @@ import (
 // A TargetStream is a single bidirectional stream between
 // the proxy and a target sansshell server
 type TargetStream struct {
+	// The parent context that owns this stream.
+	ctx context.Context
+
 	// The unique indentifier of this stream
 	streamID uint64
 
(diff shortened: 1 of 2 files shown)


proxy/server/target.go
```diff
@@ -141,8 +164,39 @@ func (s *TargetStream) Send(req proto.Message) error {
...
+	startStreamReply := &pb.ProxyReply{
+		Reply: &pb.ProxyReply_StartStreamReply{
+			StartStreamReply: &pb.StartStreamReply{
+				Target: s.target,
+				Nonce:  nonce,
+			},
+		},
+	}
+
+	grpcStream, err := s.grpcConn.NewStream(s.ctx, s.serviceMethod.StreamDesc(), s.serviceMethod.FullName())
```
sfc-gh-jchacon, line 177: you aren't locking here. Should you be?

proxy/server/target.go
```diff
@@ -512,6 +563,49 @@ func convertStatus(s *status.Status) *pb.Status {
 	}
 }
 
+// An unconnected client stream is a grpc.ClientStream
```
sfc-gh-jchacon, line 566: This feels more useful than just here in proxy. Perhaps one of the util packages?

sfc-gh-jchacon (approved):

proxy/server/target.go
sfc-gh-jallie: no. Most of this method uses the local-in-scope stream object (after associating it with the target stream to make it visible to other callers.)

_Technically_ speaking, the lock is currently unnecessary, because I don't pass back the stream ID to the caller until after the connected stream has been associated with the target, but .. putting in the mutex helps with any future cases in which we might try to access the stream prior to 'Run', and I didn't feel right about depending on the organization of the current code for threadsafety.

proxy/server/target.go
sfc-gh-jallie: per-offline chat. We'll wait until we have another potential callsite before moving this to a utility library.

PR #120, opened 2022-05-13
Instead of flags sprinkled across modules export vars from each instead.

sfc-gh-jchacon:
The in the binaries we can bind them to flags as needed.

This is a lot cleaner since flags can't conflict with each other and over time as we add modules this is bound to happen otherwise. With package vars you get a natural boundary that can't conflict.

Additionally some items will likely be duplicated too (i.e. where is the JVM binary?) and having each module attempt to create a unique flag vs one binding (flag.Func can do this for us) will be useful when scaling.

In the end this wasn't a lot of overall code change and caught a few bugs where empty flags would produce odd behavior so add some code and tests for that too.

### DIFF
--- a/auth/mtls/flags/flags.go
+++ b/auth/mtls/flags/flags.go
@@ -21,8 +21,6 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"flag"
-	"log"
 	"os"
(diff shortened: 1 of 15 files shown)


sfc-gh-jallie (approved):
LGTM

PR #122, opened 2022-05-16
Convert logfile return from fdbcli into a streaming RPC instead of unary

sfc-gh-jchacon:
Logfiles can be arbitrary sizes and we don't want unbounded memory consumption for a large file.

Instead stream them over in reasonable chunks keyed by filename.


sfc-gh-jallie (approved):
LGTM

PR #123, opened 2022-05-17
Add hooks for adding interceptors.

sfc-gh-jchacon:
Either when using the server build helper or the cmd libraries.

This way users can append their own interceptors for telemetry/etc without having to worry we're forcing a certain usage.

The only thing we force is logging and OPA. Though likely logging should be made more customizable too.


cmd/proxy-server/server/server.go
```diff
@@ -55,6 +55,18 @@ type RunState struct {
 	// entry is found. The supplied function can then do any validation it wants
 	// in order to ensure it's compliant.
 	JustificationFunc func(string) error
+	// UnaryInterceptors are any additional interceptors to be added to unary RPCs
```
sfc-gh-jallie, line 58: optional: how do you feel about the OptionFunc pattern? 

I've grown to appreciate it vs "option structs", but I realize it's a matter of taste. It does have the advantage of mirroring the option style that gRPC uses. 

we could probably refactor this into something like

```go
type runState {
  // existing runstate fields
}

type Option interface {
   apply(*runState) error
} 

type optionFunc func(*runstate) error

func (o optionFunc) apply(r *runstate) error {
  return o(r)
}

func WithLogger(l logr.Logger) Option {
  return optionFunc(func(r *runState) error {
      r.Logger = l
      return nil
  })
}

// ...other options

func (s *Server) Run(ctx context.Context, opts ...Option) {
  rs := &runState{}
  for _, o := range opts {
     o.apply(rs)
  }
 // use runstate here
}
```

sfc-gh-jallie (approved):
LGTM

cmd/proxy-server/server/server.go
sfc-gh-jchacon: I like this and agree we should go this route. Let me rework

sfc-gh-jchacon: Take another look please for sanity.

sfc-gh-jallie (approved):
I think this looks much nicer at the call sites

PR #125, opened 2022-05-17
Add in channelz service so metrics can be exposed if wanted.

sfc-gh-jchacon:


server/server.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/channelz/service"
```
sfc-gh-jallie, line 28: up to you, but 'service' seems like a likelytocollide package name. Could alias this import to 'channelz' or something likely unique

sfc-gh-jallie (approved):
LGTM

server/server.go
sfc-gh-jchacon: good point. Done

PR #128, opened 2022-05-20
Cleanup client error handling.

sfc-gh-jchacon:
In unary cases just make sure top level errors are clear these are for all targets.

In streaming receive cases track the streams we've completed. This way errors on Recv() just apply to the remaining streams, not everything.

Avoids confusion


sfc-gh-jallie (approved):
LGTM

PR #129, opened 2022-05-23
Move all connection logic inside Run() and simply do validation only up front.

sfc-gh-jchacon:
When creating a new steam all we really need to do is validate the Method and return an ID.

This allows clients to start pushing requests down before all connections are finished.

i.e. I want to health check 100 hosts with 10 timing out. I can't even start the health check until 20s have passed waiting on the dial timeout. Right now the NewStream is parallel but we still have to pay one timeout. This way no timeouts have to be paid to start work. This also allows a future dial timeout to be plumbed into the protocol since we can do blocking dials now as each will be scoped to their own go routine instead of blocking dispatcher.

This does change the contract on authz handling. Pushing authz into the TargetStream instead of at the set level.

Each target is fundamentally a discrete unit and treating this as a transaction is in hindsight (for me) a fallacy. Nothing is atomic here as any number of targets may fail for reasons. So OPA should validate each connection and be bound to each one as well. This also means you get better error messages in the end as an OPA failure is tied to the specific stream not a general error on the set. That has a bonus of being much easier to debug. Think an OPA policy which binds hosts to groups and you include one for a group you're not in. Today you get a generic set error of failing OPA without a good idea of which host triggered it, whereas with this change you'll see the specific error on the specific stream.

### DIFF
--- a/proxy/server/target.go
+++ b/proxy/server/target.go
@@ -39,14 +39,14 @@ import (
 )
 
 var (
-	// ReqChanSize is the amount of requests we'll buffer on a given stream
+	// ReqBufferSize is the amount of requests we'll buffer on a given stream
 	// while blocking to do the initial connect. After this the whole stream
 	// will block until it can proceed (or error). Exported as a var so it can
 	// be bound to a flag if wanted. By default this is only one as most RPCs
 	// are unary from the client end so a small buffer is fine. Larger numbers
 	// can cause large explosions in memory usage as potentially needing to buffer
 	// N requests per sub stream that is slow/timing out.
-	ReqChanSize = 1
+	ReqBufferSize = 1
 )
 
 // A TargetStream is a single bidirectional stream between
@@ -237,7 +237,7 @@ func (s *TargetStream) Run(nonce uint32, replyChan chan *pb.ProxyReply) {
 					if s.serviceMethod.ClientStreams() {
 						err = grpcStream.CloseSend()
 						if err != nil {
-							s.cancelFunc()
+							s.CloseWith(err)
 						}
 						return err
 					}
@@ -249,8 +249,7 @@ func (s *TargetStream) Run(nonce uint32, replyChan chan *pb.ProxyReply) {
 			authinput, err := rpcauth.NewRPCAuthInput(ctx, s.Method(), req)
 			if err != nil {
 				err = status.Errorf(codes.Internal, "error creating authz input %v", err)
-				s.errChan <- err
-				s.cancelFunc()
+				s.CloseWith(err)
 				return err
 			}
 			streamPeerInfo := s.PeerAuthInfo()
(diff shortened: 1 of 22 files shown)


proxy/server/target.go
```diff
@@ -211,13 +235,37 @@ func (s *TargetStream) Run(nonce uint32, replyChan chan *pb.ProxyReply) {
...
 					// otherwise, we're done
 					return nil
 				}
 			}
-			err := grpcStream.SendMsg(req)
+
+			authinput, err := rpcauth.NewRPCAuthInput(ctx, s.Method(), req)
+			if err != nil {
+				err = status.Errorf(codes.Internal, "error creating authz input %v", err)
+				s.errChan <- err
```
sfc-gh-jallie, line 252: you can use 'CloseWith' here,  does the same thing, but handles the non-blocking send to errChan and call to cancel(). This prevents the case in which you happen to hit multiple errors at the same time, yielding a deadlock in sending to errChan

proxy/server/target.go
```diff
@@ -211,13 +235,37 @@ func (s *TargetStream) Run(nonce uint32, replyChan chan *pb.ProxyReply) {
...
+				return err
+			}
+			streamPeerInfo := s.PeerAuthInfo()
+			authinput.Host = &rpcauth.HostAuthInput{
+				Net: streamPeerInfo.Net,
+			}
+
+			// If authz fails, close immediately with an error
+			if err := s.authorizer.Eval(ctx, authinput); err != nil {
+				s.errChan <- err
```
sfc-gh-jallie, line 263: ditto s.CloseWith() here

proxy/server/target.go
sfc-gh-jallie: or alternately, you could just return the error? errGroup cancels the ctx on the first non-err return, and should return the first error encountered from Wait(), so I'm not sure you need to write to errChan directly. Err chan is mostly for 'external' cancellation from outside the stream.

proxy/server/target.go
```diff
@@ -38,6 +38,17 @@ import (
...
 
+var (
+	// ReqChanSize is the amount of requests we'll buffer on a given stream
+	// while blocking to do the initial connect. After this the whole stream
+	// will block until it can proceed (or error). Exported as a var so it can
+	// be bound to a flag if wanted. By default this is only one as most RPCs
+	// are unary from the client end so a small buffer is fine. Larger numbers
+	// can cause large explosions in memory usage as potentially needing to buffer
+	// N requests per sub stream that is slow/timing out.
+	ReqChanSize = 1
```
sfc-gh-jallie, line 49: maybe ReqBufferSize? the fact that it's used to construct a channel is incidental.

sfc-gh-jchacon: Makes sense. Done

sfc-gh-jallie (approved):
LGTM

PR #130, opened 2022-05-23
With the proxy pushing all connection logic after stream reply we can add a dial timeout now

sfc-gh-jchacon:
Move Dial into Run() as well and plumb a timeout into here (optional).

This allow a per target dial timeout if wanted. Blocking doesn't matter since all connection logic is handled inside a per target go routine so we can safely do the separate blocking call from Dial now.

Add the ability for DialContext in TargetDialer to take additional options at calling time. This way we can slip in a WithBlock() but only in cases it's needed.

Add tests proving timeouts work and don't error in the normal case


sfc-gh-jallie (approved):
LGTM

PR #132, opened 2022-05-26
Dial timeout client side implementation

sfc-gh-jchacon:
Plumb dial options into targets/proxy. Do this by allowing an optional semi-colon after the name with a time.Duration style.

Add tests in proxy client side to validate. Also address some holes we allowed before (like blank targets).

Plug into CLI and then add integration tests to prove it all works.

### DIFF
--- a/cmd/sanssh/main.go
+++ b/cmd/sanssh/main.go
@@ -114,6 +114,23 @@ func hasPort(s string) bool {
 	return strings.LastIndex(s, "]") < strings.LastIndex(s, ":")
 }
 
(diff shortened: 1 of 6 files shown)


sfc-gh-jallie (approved):
To be honest, I'm not sure I'm a huge fan of the timeout syntax. It feels like we could probably solve 90% of the problem with a single --target-dial-timeout flag, but .. I can also see the argument for exposing as much of the underlying proto api, which does permit per-target timeouts. 

So.. tie goes to the implementor, in this case :)

PR #137, opened 2022-06-15
Implement fdbcli missing commands

sfc-gh-jchacon:
Rework CLI to be more uniform as a result (common usage, error checking). Also fix some bugs found in testing.

We now either take --exec "something" or just a bunch of args and turn that into N commands to ship over to fdbcli.

Rename help to fdbclihelp as help conflicts with the internal help (which we need to provide at the fdbcli level)

Fix tabwriter output so this actually looks sane no matter the level.


services/fdb/fdb.proto
```diff
@@ -360,6 +553,9 @@ message FDBCLITransaction {
 
 message FDBCLIUnknownAction {}
 
+// FDBCLIRequest defines the fdbcli command(s) to run.
+// This will replicate fdbcli --exec "command(s)" (where N commands are
+// separated by ;) NEXT: 19 DO NOT USE: 11, 12, 13
```
sfc-gh-jallie, line 558: you can mark 11,12,13 as 'reserved' to prevent re-use: https://developers.google.com/protocol-buffers/docs/proto3#reserved

services/fdb/fdb.proto
```diff
@@ -360,6 +553,9 @@ message FDBCLITransaction {
 
 message FDBCLIUnknownAction {}
 
+// FDBCLIRequest defines the fdbcli command(s) to run.
+// This will replicate fdbcli --exec "command(s)" (where N commands are
```
sfc-gh-jallie, line 557: does --exec foo;bar provide some transactional/sematic difference vs 'fdb foo ; fdb bar' (e.g. would it be different than running them in a shell script sequentially?)

services/fdb/server/fdbcli.go
```diff
@@ -865,13 +1064,44 @@ func parseFDBCLISleep(req *pb.FDBCLISleep) ([]string, []captureLogs, error) {
...
 	return args, nil, nil
 }
 
+func parseFDBCLISuspend(req *pb.FDBCLISuspend) ([]string, []captureLogs, error) {
+	args := []string{"suspend"}
+
+	if req.Request == nil {
+		return nil, nil, status.Error(codes.InvalidArgument, "suspend requires request to be filled in")
+	}
+	switch req.Request.(type) {
```
sfc-gh-jallie, line 1091: do you want to return an error in the 'default' case here? might make it more obvious to debug if we added a new unhandled case

services/fdb/client/client.go
```diff
@@ -298,36 +311,83 @@ func (r *fdbCLICmd) SetFlags(f *flag.FlagSet) {
...
+	c = setupFDBCLI(f)
+
 	args = append(args, r.req)
-	return c.Execute(ctx, args...)
+	var cmd []string
+
+	// Join everything back into one string and then resplit. So a command of fdbcli "kill; kill X" will
+	// parse through here one by one.
+	t := strings.Join(f.Args(), " ")
+	for _, a := range strings.Split(t, " ") {
```
sfc-gh-jallie, line 348: could you split on ";" and strings.Trim() the result of leading/trailing spaces?

services/fdb/fdb.proto
sfc-gh-jchacon: Done

services/fdb/client/client.go
sfc-gh-jchacon: Yes, it does simplify this a lot in the end

sfc-gh-jallie (approved):
LGTM

PR #138, opened 2022-06-15
Add some precommit checks which cover things integration is doing

sfc-gh-jchacon:
Also add some linting we could use.

Go through the lints and fix the obvious ones. It definitely found some bugs.


sfc-gh-jallie (approved):
LGTM

PR #139, opened 2022-06-21
Add debugging support for cert reloads

sfc-gh-jchacon:
Caching a context is a bad idea generally but we should pull any logger out for the incoming context and keep that in the wrapped Credential. This way we can log later and construct a logging context when we have no context except Background() to pass along.

Add logging around reload logic.

Technically a given credential can happen in N threads so anything writing here needs locking around it so add a mutex protecting access to it as we could panic if it's swapped while another thread is using it. Add some tests to double check this also.

### DIFF
--- a/auth/mtls/client.go
+++ b/auth/mtls/client.go
@@ -22,12 +22,14 @@ import (
 	"crypto/x509"
 	"fmt"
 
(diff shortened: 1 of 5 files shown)


sfc-gh-jallie (approved):
LGTM

PR #140, opened 2022-06-22
The init needs to register a cserver in the conf code or we end up registering the  fdbcli code twice (which is fatal).

sfc-gh-jchacon:


sfc-gh-jallie (approved):
LGTM

PR #143, opened 2022-06-28
Fix bidi stream blocking naive proxy setup causes.

sfc-gh-jchacon:
There was an assumption we could blast N start streams down and then loop reading responses.

In practice this works until about 2k starts. At that point the remote server has sent enough responses that it will now block sending back.

The server requires the client to clear its response queue as it won't buffer infinitely to avoid scaling issues.

As a result we deadlock around ~2k+ start streams and hang until the context expires (which does properly unwind all of this and the last recv shows a context expiring).

Solve this by moving Send/Recv into separate go routines since the state they are modifying is disjoint. We need to send N starts and then we're done. We need to get back N replies but they aren't correlated past that.

Add some logging at the end so debugging is easier since Send can return EOF and means "the error is on the Recv side".


sfc-gh-jallie (approved):
LGTM

PR #147, opened 2022-07-06
Add the ability to use the helper functions for starting proxy/server but also register your own services.

sfc-gh-jchacon:
Create option variants to register arbitrary RPC services against the RPC server that's created.

Move channelz/reflection/etc up to the main() function. This way users aren't forced to inherit these
services unless they want them. Also means users can add their own arbitrary RPC services
to the server as well.

### DIFF
--- a/cmd/proxy-server/server/server.go
+++ b/cmd/proxy-server/server/server.go
@@ -28,12 +28,9 @@ import (
 	"github.com/Snowflake-Labs/sansshell/auth/mtls"
 	"github.com/Snowflake-Labs/sansshell/auth/opa/rpcauth"
 	"github.com/Snowflake-Labs/sansshell/proxy/server"
-	ss "github.com/Snowflake-Labs/sansshell/services/sansshell/server"
 	"github.com/Snowflake-Labs/sansshell/telemetry"
 	"github.com/go-logr/logr"
 	"google.golang.org/grpc"
-	channelz "google.golang.org/grpc/channelz/service"
-	"google.golang.org/grpc/reflection"
 )
 
 // runState encapsulates all of the variable state needed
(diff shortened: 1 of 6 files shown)


cmd/proxy-server/server/server.go
```diff
@@ -174,6 +172,13 @@ func WithAuthzHook(hook rpcauth.RPCAuthzHook) Option {
 	})
 }
 
+func WithAdditionalRPCService(s func(*grpc.Server)) Option {
```
sfc-gh-jallie, line 175: thinking out loud.., feel free to leave as-is since you may have considered all of this. 

It originally struck me as odd that this method took a 'func that wants a grpc.Server', which felt like a pretty broad interface that could be used to do anything arbitrary to the server (not just register services), and I wondered why it didn't take a 'ServiceRegistrar', but then noticed that reflection wanted a couple of extra methods from server, and had it's own interface definition.

I'd also wondered whether it was 'weird' that we had effectively 2 different ways to add services (_ import + this) and whether we could collapse them, to avoid confusion for people trying to add new service modules. (I'm not sure we really could, at least not gracefully).

As a half-way measure, we could do something like:

```
func WithServerReflection() Option {
   return optionFunc(func(r *runState) error {
      r.services  = append(r.services, func(s *grpc.Server) {
          reflection.Register(s)
     })
  }
}

func WithChannelz Option {
  ... 
}
```

to still leave these as user-toggles without needing something quite so broad


(Another option would be to make it clear that this was a broad 'escape hatch' API to directly access the server, and just call it `WithRawServerOption`)

sfc-gh-jallie (approved):
LGTM

cmd/proxy-server/server/server.go
sfc-gh-jchacon: I went with WithRawServerOption as it makes more sense. Granted you can generally only serve/lookup registration and register. I thought about just exposing it as a register but it's more complicated since some packages like reflection wrap their object, etc.

PR #151, opened 2022-07-16
Fix JSON logging of RPCAuthInput

sfc-gh-jchacon:
Restructure how we log an RPCAuthInput.

We can't just json marshal and then stringify it into the logs. It gets escaped then which makes it not workable if you assume this entry is JSON (which it should be).

To get the 2 fields which are already marshaled we'd like to read those as logr will print slices as [byte, byte, ...] which isn't useful.

So maintain a mirror type of RPCAuthInput where the 2 marshal'd fields are strings and then copy everything over and return that struct for logging purposes.

It does mean input.message is escaped still but that's easier than a byte stream and the rest of the log is fully parseable JSON now.

Long term we should likely write our own logr implementation where we can disable escaping on a per field basis.


sfc-gh-jallie (approved):
LGTM

PR #153, opened 2022-07-22
Fix ordering for yum commands.

sfc-gh-jchacon:
disablerepo must come first or it'll override any enablerepo you might have just set.

Generally people want --disablerepo=* --enablerepo=REPO to just
process a specific repo and reversing makes that impossible.


sfc-gh-jallie (approved):
LGTM

PR #154, opened 2022-08-01
Add a field to RPCAuthInput indicating whether this was generated by the proxy.

sfc-gh-jchacon:
If false we assume it's direct RPC's (via the interceptor).

This allows hooks/code/policy to make separate decisions based on the origin of the auth request.

### DIFF
--- a/auth/opa/rpcauth/input.go
+++ b/auth/opa/rpcauth/input.go
@@ -52,12 +52,20 @@ type RPCAuthInput struct {
 	// Information about the host serving the RPC.
 	Host *HostAuthInput `json:"host"`
 
-	// Whether this RPCAuthInput was generated via a proxy
-	// or directly in the RPC interceptor.
-	FromProxy bool `json:"from_proxy"`
+	// Information about the environment in which the policy evaluation is
+	// happening.
+	Environment *EnvironmentInput `json:"environment"`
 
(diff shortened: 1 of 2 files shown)


auth/opa/rpcauth/input.go
```diff
@@ -54,6 +54,10 @@ type RPCAuthInput struct {
 	// Information about the host serving the RPC.
 	Host *HostAuthInput `json:"host"`
 
+	// Whether this RPCAuthInput was generated via a proxy
+	// or directly in the RPC interceptor.
+	FromProxy bool `json:"from_proxy"`
```
sfc-gh-jallie, line 59: optional:

Assuming we can't stuff this into 'Extensions', it might be worth stuffing this into a more general field that could capture other things, rather than as a top-level.

Something like:


```go
// EnvironmentInput contains information about the environment in which the policy evaluation is
// happening.
type EnvironmentInput struct {
   // True if the policy evaluation is being performed by some entity other than the host
   // (for example, policy checks performed by an authorizing proxy or other agent)
   NonHostPolicyCheck bool `json:"non_host_policy_check"`
}
```

sfc-gh-jchacon: Good point. I like this. Integrated

PR #155, opened 2022-08-02
After more consideration revert the logging on rpcauth.

sfc-gh-jchacon:
Instead just let the logger convert the struct based on tags, etc. Default logr/funcr will print message as a byte array. But..this then allows someone to write their own logger to convert it.

Forcing it to a string ends up having side effects in logr/funcr otherwise.

Cleanup a debug log that slipped into fdbcli


sfc-gh-jallie (approved):
LGTM

PR #157, opened 2022-08-09
Things calling exec directly not handling bad commands (bad path)

sfc-gh-jchacon:
If we exec something that returns non-zero it can happen for 2 reasons:

1. Command simply ran and exited non-zero. That works as expected mostly except the client code didn't error out for any target exiting this way. So fix client code to exit non-zero if any target fails.

2. Command couldn't run (i.e. exec error). That's returned from exec.RunCommand as an error. But we may have nothing in stderr. In this case we should copy the error to stderr since errors here are for commands not being setup correctly. Exit code was already right but otherwise we ate the error string and couldn't see why it failed. 

Adjust all the tests so they account for this too and add an integration test.

Ansible had same issues as exec due to this.

### DIFF
--- a/services/ansible/server/ansible_test.go
+++ b/services/ansible/server/ansible_test.go
@@ -103,6 +103,7 @@ func TestRun(t *testing.T) {
 		returnCodeNonZero bool
 		stdout            string
 		stderr            string
+		stderrNocmp       bool
 	}{
 		{
 			name:    "A non-absolute bin path",
@@ -119,6 +120,7 @@ func TestRun(t *testing.T) {
 			bin:               "/non-existant-command",
 			path:              path,
 			returnCodeNonZero: true,
+			stderrNocmp:       true, // can't guarentee error message cross platform
 		},
 		{
 			name:              "A command that exits non-zero",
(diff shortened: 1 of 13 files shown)


services/ansible/server/ansible_test.go
```diff
@@ -119,6 +120,7 @@ func TestRun(t *testing.T) {
 			bin:               "/non-existant-command",
 			path:              path,
 			returnCodeNonZero: true,
+			stderrNocmp:       true, // can't guarentee error message cross platform
```
sfc-gh-jallie, line 123: nit: typo: "guarantee"

services/util/util.go
```diff
@@ -250,7 +250,15 @@ func RunCommand(ctx context.Context, bin string, args []string, opts ...Option)
 	logger.Info("executing local command", "cmd", cmd.String())
 	run.Error = cmd.Run()
```
sfc-gh-jallie, line 251: I think we could also switch to using "Start()" and "Wait()", which return distinct errors for "couldn't execute the command" and "Command exited", but this is fine

sfc-gh-jallie (approved):
LGTM

services/ansible/server/ansible_test.go
sfc-gh-jchacon: Removed in the end

services/util/util.go
sfc-gh-jchacon: So...turns out that is equivalent (all Run() does is call Start()+Wait() effectively) so errors come out the same.

But...Thinking about it an ExitError isn't really an error in this API since we provide exit code and ExitError guarantees the stdout/stderr output is there if generated. So we shouldn't return an error for that case and nil it out.
It simplifies handling in exec/ansible but did add some work in process/fdbcli/packages.

PTAL

sfc-gh-jallie (approved):
LGTM

PR #158, opened 2022-08-19
Add batch support to CLI.

sfc-gh-jchacon:
If we have thousands of targets at some point the proxy will reach a CPU/RAM limit. For most operations this isn't needed and we could instead just send N batches.

If batch > 0 break up the target list into N chunks of batch size and send them over.

Adjust integration tests to force a batch size of 1 to validate. Turn back off for tail/ptrace/core where we expect 2 running in parallel for testing.


cmd/sanssh/client/client.go
```diff
@@ -267,6 +257,61 @@ func Run(ctx context.Context, rs RunState) {
...
+	// How many batches? Integer math truncates so we have to do one more after for remainder.
+	for i := 0; i < batchCnt; i++ {
+		// Set up a connection to the sansshell-server (possibly via proxy).
+		conn, err := proxy.Dial(rs.Proxy, rs.Targets[i*rs.BatchSize:rs.BatchSize*(i+1)], ops...)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Could not connect to proxy %q node(s) in batch %d: %v\n", rs.Proxy, i, err)
+			os.Exit(1)
+		}
+		defer func() {
+			if err := conn.Close(); err != nil {
```
sfc-gh-jallie, line 283: probably not a big deal, but defer is function scoped, so you won't close any connections until you've processed all batches and exit Run(). (We already have this with open files, but in this case, you're consuming targets/batchSize proxy conns, so it _could_ be meangingful in the case of many targets and a tiny batch size)

in this case, the scope of conn is pretty small, and there are no early returns, so you could probably just explicitly close it after line 292 ( although someone could later introduce a bug with an early exit before then).

Up to you.

sfc-gh-jallie (approved):
LGTM

cmd/sanssh/client/client.go
sfc-gh-jchacon: True. Just closed inline.

PR #160, opened 2022-09-01
Move golangci-lint into a separate action outside of pre-commit.

sfc-gh-jchacon:
This way it's pre-canned and not installing each time from the internet.


sfc-gh-jallie (approved):
LGTM

PR #161, opened 2022-09-02
add another hook and flag to set tls.Config

sfc-gh-tmerz:
add another hook and flag to set tlc.Config directly rather than relying on credsSource


cmd/sansshell-server/server/server.go
```diff
@@ -192,3 +198,22 @@ func Run(ctx context.Context, opts ...Option) {
...
+	var creds credentials.TransportCredentials
+	var err error
+	if rs.credSource != "" && rs.tlsConfig != nil {
+		rs.logger.Error(nil, "both credSource and tlsConfig are defined", "credSource", rs.credSource)
+		os.Exit(1)
+	} else if rs.credSource != "" {
+		creds, err = mtls.LoadServerCredentials(ctx, rs.credSource)
+		if err != nil {
+			rs.logger.Error(err, "mtls.LoadServerCredentials", "credsource", rs.credSource)
+			os.Exit(1)
```
sfc-gh-jchacon, line 213: Make this function simply return an error and then have Run() log it and exit. That way all exit handling is in one place.

cmd/sansshell-server/server/server.go
```diff
@@ -73,6 +76,14 @@ func WithPolicy(policy string) Option {
 	})
 }
 
+// WithTlsConfig applies a supplied tls.Config object to the gRPC server.
```
sfc-gh-jchacon, line 79: This isn't the only place. Also need to do cmd/proxy-server/server/server.go since it does a similar thing (this wrapper is only for the sansshell-server part)

cmd/sansshell-server/server/server.go
sfc-gh-tmerz: ok, thanks! I was not sure what the best practices were for error handling like that.

cmd/sansshell-server/server/server.go
sfc-gh-tmerz: Thanks for pointing that out. I think that I have gotten the other places as well.

cmd/proxy-server/server/server.go
```diff
@@ -297,3 +312,37 @@ func Run(ctx context.Context, opts ...Option) {
 		os.Exit(1)
 	}
 }
+
+// extracts transport credentials from runState. Will error if both credSource and tlsConfig are both specified
```
sfc-gh-jchacon, line 316: Use godoc style here:

extractClientTransportCredentialsFromRunState extracts transport ...

cmd/proxy-server/server/server.go
```diff
@@ -297,3 +312,37 @@ func Run(ctx context.Context, opts ...Option) {
...
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		creds = credentials.NewTLS(rs.tlsConfig)
+	}
+	return creds, nil
+}
+
+// extracts transport credentials from runState. Will error if both credSource and tlsConfig are both specified
```
sfc-gh-jchacon, line 333: use godoc style here

cmd/proxy-server/server/server.go
```diff
@@ -297,3 +312,37 @@ func Run(ctx context.Context, opts ...Option) {
...
 	}
 }
+
+// extracts transport credentials from runState. Will error if both credSource and tlsConfig are both specified
+func extractClientTransportCredentialsFromRunState(ctx context.Context, rs *runState) (credentials.TransportCredentials, error) {
+	var creds credentials.TransportCredentials
+	var err error
+	if rs.credSource != "" && rs.tlsConfig != nil {
+		return nil, fmt.Errorf("both credSource and tlsConfig are defined for the client")
+	} else if rs.credSource != "" {
```
sfc-gh-jchacon, line 322: Drop the else and outdent this if as the above case is terminal if it's taken (no need for the else then)

cmd/proxy-server/server/server.go
```diff
@@ -297,3 +312,37 @@ func Run(ctx context.Context, opts ...Option) {
...
+	return creds, nil
+}
+
+// extracts transport credentials from runState. Will error if both credSource and tlsConfig are both specified
+func extractServerTransportCredentialsFromRunState(ctx context.Context, rs *runState) (credentials.TransportCredentials, error) {
+	var creds credentials.TransportCredentials
+	var err error
+	if rs.credSource != "" && rs.tlsConfig != nil {
+		return nil, fmt.Errorf("both credSource and tlsConfig are defined for the server")
+	} else if rs.credSource != "" {
```
sfc-gh-jchacon, line 339: same here. drop else and outdent

cmd/sansshell-server/server/server.go
```diff
@@ -193,3 +204,20 @@ func Run(ctx context.Context, opts ...Option) {
...
 	}
 }
+
+// extracts transport credentials from runState. Will error if both credSource and tlsConfig are both specified
+func extractTransportCredentialsFromRunState(ctx context.Context, rs *runState) (credentials.TransportCredentials, error) {
+	var creds credentials.TransportCredentials
+	var err error
+	if rs.credSource != "" && rs.tlsConfig != nil {
+		return nil, fmt.Errorf("both credSource and tlsConfig are defined")
+	} else if rs.credSource != "" {
```
sfc-gh-jchacon, line 214: same, drop else and outdent

cmd/sansshell-server/server/server.go
```diff
@@ -193,3 +204,20 @@ func Run(ctx context.Context, opts ...Option) {
 		os.Exit(1)
 	}
 }
+
+// extracts transport credentials from runState. Will error if both credSource and tlsConfig are both specified
+func extractTransportCredentialsFromRunState(ctx context.Context, rs *runState) (credentials.TransportCredentials, error) {
```
sfc-gh-jchacon, line 209: use godoc style

sfc-gh-jchacon (approved):

PR #164, opened 2022-09-23
Fix RecvMsg in the single target case.

sfc-gh-jchacon:
We need to allow this to be called via Recv() as well which means handling proto.Message at the return type too.

So check for both types and base this on # targets for which is valid (or not).

Write a test which catches this case since we missed it before.


sfc-gh-jallie (approved):
LGTM

PR #169, opened 2022-10-11
If Ansible returns an error from RunCommand handle it correctly.

sfc-gh-jchacon:
Before we checked but then returned err which we know is nil because it got through the check above.

This means we passed nil, nil back to grpc which doesn't like it:

2022/10/11 00:18:29 ERROR: [core] [Server #4] grpc: server failed to encode response: rpc error: code = Internal desc = grpc: error while marshaling: proto: Marshal called with nil

Instead capture the error and put it inside a proper status.Error


sfc-gh-jallie (approved):
LGTM

PR #171, opened 2022-10-12
Add service enable/disable support and tests

sfc-gh-jchacon:
Enable/disable does an implicit reload when you do this via systemctl so that has to be done as a second step.

Otherwise mostly looks like other actions except return type is more direct than the channel.


services/service/server/server_linux.go
```diff
@@ -266,10 +273,20 @@ func (s *server) Action(ctx context.Context, req *pb.ActionRequest) (*pb.ActionR
 	// NB: delivery of a value on resultchan respects context cancellation, and will
 	// deliver a value of 'cancelled' if the ctx is cancelled by a client disconnect,
 	// so it's safe to do a simple recv.
-	result := <-resultChan
-	if result != operationResultDone {
-		return nil, status.Errorf(codes.Internal, "error performing action %v: %v", req.Action, result)
+	// Enable/disable don't use this method so we skip the channel (since it would hang)
+	// and instead force a reload which is what systemctl does when it enables/disables.
+	switch req.Action {
```
sfc-gh-jallie, line 278: nit/optional: I typically always add a `default`: case that errors, in case someone adds a new enum value which isn't handled (there is one above, so it would already, but in case that value also needs handling, it's nice to have it be explicit)

sfc-gh-jallie (approved):
LGTM

services/service/server/server_linux.go
sfc-gh-jchacon: True. Added

PR #174, opened 2022-10-18
Add package cleanup command.

sfc-gh-jchacon:
In some cases a YUM operation can leave things partially finished. i.e. doing a live update of the sansshell YUM package could result in a server restart before yum has completed.

Allow this to fix via Cleanup which for YUM implements yum-complete-transaction --cleanup-only

Add client support and tests


sfc-gh-jallie (approved):
LGTM

PR #175, opened 2022-11-05
Adds convenience command to issue DNS queries

### LINKED ISSUE #167: Add DNS lookup features
It would be useful to be able to perform DNS queries from the perspective of hosts in production running `sansshell`, in order to debug e.g. private DNS zone issues inside cloud providers, or host DNS configuration issues.

Ideally any implementation would use the system resolver (i.e. `libresolv`, `/etc/resolv.conf` etc etc.). I understand the `net/dns` in Go does this by default, so hopefully will not be too difficult.

elsesiy:
Adds a new subpackage 'dns' with a Lookup command that simply takes a hostname and returns the associated IPs for which we're using the default resolver.

> The method for resolving domain names, whether indirectly with functions like Dial or directly with functions like LookupHost and LookupAddr, varies by operating system.
>
> On Unix systems, the resolver has two options for resolving names. It can use a pure Go resolver that sends DNS requests directly to the servers listed in /etc/resolv.conf, or it can use a cgo-based resolver that calls C library routines such as getaddrinfo and getnameinfo.
>
> By default the pure Go resolver is used, because a blocked DNS request consumes only a goroutine, while a blocked C call consumes an operating system thread. When cgo is available, the cgo-based resolver is used instead under a variety of conditions: on systems that do not let programs make direct DNS requests (OS X), when the LOCALDOMAIN environment variable is present (even if empty), when the RES_OPTIONS or HOSTALIASES environment variable is non-empty, when the ASR_CONFIG environment variable is non-empty (OpenBSD only), when /etc/resolv.conf or /etc/nsswitch.conf specify the use of features that the Go resolver does not implement, and when the name being looked up ends in .local or is an mDNS name.

Source: https://pkg.go.dev/net#hdr-Name_Resolution

Fix #167


elsesiy:
Server:
```
./sansshell-server
2022/11/05 23:33:09 util.go:51: sanshell-server: "level"=0 "msg"="using built-in policy"
2022/11/05 23:33:09 server.go:65: sanshell-server: "level"=0 "msg"="loading new server cert"
2022/11/05 23:33:09 server.go:70: sanshell-server: "level"=0 "msg"="loaded new server cert" "error"=null
2022/11/05 23:33:23 telemetry.go:162: sanshell-server: "level"=0 "msg"="new request" "method"="/Dns.Lookup/Lookup" "peer-address"="127.0.0.1:54975"
2022/11/05 23:33:23 dns.go:41: sanshell-server: "level"=0 "msg"="dns request" "method"="/Dns.Lookup/Lookup" "peer-address"="127.0.0.1:54975" "hostname"="app.snowflake.com"
```

Client:
```
./sanssh --targets=localhost dns lookup app.snowflake.com
44.242.126.255
35.80.239.15
34.210.48.230
```

elsesiy:
@sfc-gh-jchacon @sfc-gh-msuriar I took a stab at #167, PTAL

services/dns/client/client.go
```diff
@@ -0,0 +1,104 @@
...
+	return `lookup:
+    Performs a simple DNS lookup for the provided hostname on the remote server.
+`
+}
+
+func (p *lookupCmd) SetFlags(f *flag.FlagSet) {}
+
+func (p *lookupCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interface{}) subcommands.ExitStatus {
+	state := args[0].(*util.ExecuteState)
+	if f.NArg() == 0 {
```
sfc-gh-jchacon, line 75: if f.Narg() != 1

i.e. this catches people adding extra things that don't parse then too

services/dns/client/client.go
```diff
@@ -0,0 +1,104 @@
...
+		fmt.Fprintln(os.Stderr, "Please specify a hostname to lookup.")
+		return subcommands.ExitUsageError
+	}
+
+	c := pb.NewLookupClientProxy(state.Conn)
+
+	hostname := f.Args()[0]
+	req := &pb.LookupRequest{
+		Hostname: hostname,
```
sfc-gh-jchacon, line 84: Hostname: f.Arg(0)

and just eliminate the hostname var (it's not used anywhere else)

services/dns/dns.proto
```diff
@@ -0,0 +1,38 @@
...
+
+// LookupRequest describes the DNS query request
+message LookupRequest {
+  // The hostname to query
+  string hostname = 1;
+}
+
+// LookupReply describes the result of a DNS query
+message LookupReply {
+  bytes result = 1;
```
sfc-gh-jchacon, line 37: why not a string or if this is N of them a repeated string?

services/dns/server/dns.go
```diff
@@ -0,0 +1,84 @@
...
+	pb "github.com/Snowflake-Labs/sansshell/services/dns"
+)
+
+var (
+	// Create package level resolver such that it can be replaced during testing
+	defaultResolver = func() Resolver {
+		return resolver{}
+	}
+)
+
```
sfc-gh-jchacon, line 40: This seems a bit overly complicated for a test fake.

You can just do:

var (
  resolver = net.DefaultResolver.LookupIP
)

and use it directly below and then in tests just replace that with a function that matches that signature. No need then for an interface and a struct, impl, etc.

services/dns/server/dns.go
```diff
@@ -0,0 +1,84 @@
...
+	logger.Info("dns request", "hostname", hostname)
+	// TODO(elsesiy): We only care about ipv4 for now but we could allow clients to explicitly specify opts such as network, prefer go resolver, etc.
+	ips, err := defaultResolver().LookupIP(ctx, "ip4", hostname)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to lookup %q", hostname)
+	}
+
+	var outString string
+	for _, ip := range ips {
+		outString += fmt.Sprintf("%s\n", ip.String())
```
sfc-gh-jchacon, line 67: Ok..based on my proto comments this should just be a repeated string field and copy each returned IP as an entry. Then the client can process how they want. This is more complicated since a client who might want to iterate across now has to do string parsing to separate the entries.

services/dns/client/client.go
elsesiy: done in 3f31c60002e0aba6c8a2ae49ccabc7211d1db494

services/dns/dns.proto
elsesiy: Initially I thought about the `dig` output format but it's much cleaner to not fiddle around with a byte array here, changed in 3f31c60002e0aba6c8a2ae49ccabc7211d1db494

services/dns/client/client.go
elsesiy: done in 3f31c60002e0aba6c8a2ae49ccabc7211d1db494

services/dns/server/dns.go
elsesiy: done in 3f31c60002e0aba6c8a2ae49ccabc7211d1db494

elsesiy:
Slight output format change unless we want to preserve the trailing newline
```
./sansshell-server
2022/11/09 14:15:09 util.go:51: sanshell-server: "level"=0 "msg"="using built-in policy"
2022/11/09 14:15:09 server.go:65: sanshell-server: "level"=0 "msg"="loading new server cert"
2022/11/09 14:15:09 server.go:70: sanshell-server: "level"=0 "msg"="loaded new server cert" "error"=null
2022/11/09 14:15:20 telemetry.go:162: sanshell-server: "level"=0 "msg"="new request" "method"="/Dns.Lookup/Lookup" "peer-address"="127.0.0.1:61584"
2022/11/09 14:15:20 dns.go:45: sanshell-server: "level"=0 "msg"="dns request" "method"="/Dns.Lookup/Lookup" "peer-address"="127.0.0.1:61584" "hostname"="app.snowflake.com"

./sanssh --targets=localhost dns lookup app.snowflake.com
44.242.126.255
34.210.48.230
35.80.239.15%  <-- no new line anymore
```

services/dns/server/dns_test.go
```diff
@@ -0,0 +1,95 @@
...
+		want    []string
+		wantErr bool
+	}{
+		"dns lookup succeeds": {testee: "gotest.com", want: []string{"1.3.3.7"}, wantErr: false},
+		"invalid hostname":    {testee: "gotest", wantErr: true},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			resolver = mockResolver{shouldFail: tc.wantErr}.LookupIP
```
sfc-gh-jchacon, line 79: For cleanliness you should do:

origResolver := resolver
t.Cleanup(func () {
  resolver = origResolver
})
resolver = mockResolver{shouldFail: tc.wantErr}.LookupIP

This way each test invocation is hermetic and leaves things as it found them and you don't get side effects between tests (not today but could later if we add more options)

services/dns/client/client.go
```diff
@@ -0,0 +1,104 @@
...
+			fmt.Fprintf(e, "All targets - could not execute: %v\n", err)
+		}
+		return subcommands.ExitFailure
+	}
+	for r := range resp {
+		if r.Error != nil {
+			fmt.Fprintf(state.Err[r.Index], "Dns lookup failure for target %s (%d) - error - %v\n", r.Target, r.Index, r.Error)
+			continue
+		}
+		fmt.Fprint(state.Out[r.Index], strings.Join(r.Resp.Result, "\n"))
```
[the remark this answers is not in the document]
elsesiy: I addressed all comments in c31c74b0d31126fbb7d94d3937c39478d951bec6, just wasn't sure what the preferred output format is. Thank you!

PR #179, opened 2023-02-11
Add a mermaid diagram to README.md

sfc-gh-srhodes:
The diagram helps give a quick visualization of the structure of a sansshell deployment, so I think it's useful to include


sfc-gh-ajoyner (approved):
Thanks for a lovely diagram, Steven!

PR #186, opened 2023-02-25
Add a way to do command line completion in sanssh.

sfc-gh-srhodes:
See steps in the readme for how to enable this. It relies on invoking sanssh for suggestions. Downstream clients can use `client.AddCommandLineCompletion` to take advantage of this.

This completion automatically registers subcommand packages and their flags. To register subcommands of subcommands, the parent subcommand needs to implement HasSubpackage. I've updated all the services to implement that interface.

There's also a PredictArgs interface that can be used to predict command line arguments of subcommands. I've experimented with adding this to some of the fdb subcommands.

Finally, I've tweaked the default fail-early behavior so that there's help text from running a command like `sanssh` or `sanssh exec` insted of a cryptic "Must set a target or a proxy" error.

### DIFF
--- a/cmd/sanssh/client/completion.go
+++ b/cmd/sanssh/client/completion.go
@@ -94,7 +94,7 @@ func (c *cmdCompleter) GetFlag(flag, prefix string) []string {
 	if predictor := c.flagPredictions[flag]; predictor != nil {
 		return predictor(prefix)
 	}
-	return []string{prefix}
+	return nil
 }
 func (c *cmdCompleter) GetArgs(prefix string) []string {
 	return nil
@@ -139,7 +139,7 @@ func (c *subCompleter) ListFlags() []string {
 }
 
 func (c *subCompleter) GetFlag(flag, prefix string) []string {
-	return []string{prefix}
+	return nil
 }
 
 func (c *subCompleter) IsBoolFlag(fl string) bool {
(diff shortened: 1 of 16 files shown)


cmd/sanssh/client/completion.go
```diff
@@ -0,0 +1,227 @@
...
+type Predictor func(prefix string) []string
+
+// AddCommandLineCompletion adds command line completion for shells. It should be called after
+// registering all subcommands and before calling flag.Parse().
+//
+// topLevelFlagPredictions gives optional finer-grained control over predictions on the top-level
+// flags.
+func AddCommandLineCompletion(topLevelFlagPredictions map[string]Predictor) {
+	// If COMP_LINE is present, we're running as command completion.
+	compLine, ok := os.LookupEnv("COMP_LINE")
```
sfc-gh-jallie, line 210: out of curiosity, would it be possible to implement this in the same basic way that 'help' and 'commands' work in 'subcommands' (i.e. as just another subcommand)?

that is, instead of `COMP_LINE="--" ./sanssh file foo`, could you implement it as `./sanssh complete <args>`

sfc-gh-srhodes: It'd require writing a bit of bash that I'd like to avoid. COMP_LINE is one of the defined variables for the `complete` command, so avoiding the environmental variable here would likely mean creating a bash function that invokes `./sanssh complete $COMP_LINE`.

I've been reading through "Programmable Completion" in https://linux.die.net/man/1/bash to figure out what's possible.

sfc-gh-srhodes: Resolving because I like the environmental variable more than a bash wrapper.

sfc-gh-jallie (approved):
LGTM

PR #188, opened 2023-02-28
Add README badges and update CI to run continuously.

sfc-gh-srhodes:
This makes the README slightly friendlier. Badges are somewhat copied from https://github.com/snowflakedb/gosnowflake.

The build status won't be valid until after submission because it'll only start having results post-submission. I've now made builds happen daily and on every update to main in addition to on PRs.


sfc-gh-jallie (approved):
LGTM <3

PR #189, opened 2023-03-03
Allow loading roots of trust from a directory.

sfc-gh-srhodes:
Snowflake has an internal use case where we have multiple root .pem files in a single directory. We currently concatenate together all the files for sanshell, but avoiding the concatenation makes the system slightly simpler.

We'd previously returned an error when asking for a directory, so this change makes us slightly more permissive.

### DIFF
--- a/auth/mtls/common.go
+++ b/auth/mtls/common.go
@@ -30,37 +30,31 @@ import (
 // If the file is a directory, LoadRootOfTrust will load all files
 // in that directory.
 func LoadRootOfTrust(path string) (*x509.CertPool, error) {
-	capool := x509.NewCertPool()
-
-	loadRoot := func(filename string) error {
-		ca, err := os.ReadFile(filename)
-		if err != nil {
-			return fmt.Errorf("could not read %q: %w", filename, err)
-		}
-		if !capool.AppendCertsFromPEM(ca) {
-			return fmt.Errorf("could not add CA cert from %q to pool: %w", filename, err)
-		}
-		return nil
-	}
-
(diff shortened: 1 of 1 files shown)


auth/mtls/common.go
```diff
@@ -20,21 +20,48 @@ import (
...
 // file.
-func LoadRootOfTrust(filename string) (*x509.CertPool, error) {
-	// Read in the root of trust for client identities
-	ca, err := os.ReadFile(filename)
+// If the file is a directory, LoadRootOfTrust will load all files
+// in that directory.
+func LoadRootOfTrust(path string) (*x509.CertPool, error) {
+	capool := x509.NewCertPool()
+
+	loadRoot := func(filename string) error {
```
sfc-gh-jallie, line 35: optional: feel free to ignore

you could structure this a bit differently and avoid the need for the function var

```
var certfiles []string
fi, err := os.Stat(path)
if err != nil { return ... }

if fi.IsDir() {
 files, err := os.ReadDir(path)
 if err != nil { return .. }
 for _, f := range files {
    certfiles = append(certfiles, filepath.join(path, f.Name())
 }
} else {
 certfiles = []string{path}
}

pool := x509.NewCertPool()
for _, file := range certfiles {
   // read file
  // append to pool
}
```

sfc-gh-jallie (approved):
LGTM

auth/mtls/common.go
sfc-gh-srhodes: I like it, it's simpler. Changed.

PR #190, opened 2023-03-03
Add an option for an extremely minimal web debug server

sfc-gh-srhodes:
server.WithDebugPort now allows you to start a debug port in both sansshell-server and proxy-server, currently with three things.

- Trivial index page
- /debug/pprof/ and children
- /metrics

This expands the public interface of this package, so some caution is warranted. Some notable callouts:

- There's no exposed interfaces for editing what gets displayed or adding your own pages
- There's no way to reduce the number of things being displayed if someone wants to make the debug handler less revealing

We can add more Options in the future and there's always the escape hatch of someone creating their own http server, so that seems like an acceptable tradeoff.

### DIFF
--- a/cmd/proxy-server/main.go
+++ b/cmd/proxy-server/main.go
@@ -64,7 +64,7 @@ var (
 	clientPolicyFlag = flag.String("client-policy", "", "OPA policy for outbound client actions (i.e. connecting to sansshell servers). If empty no policy is applied.")
 	clientPolicyFile = flag.String("client-policy-file", "", "Path to a file with a client OPA.  If empty uses --client-policy")
 	hostport         = flag.String("hostport", "localhost:50043", "Where to listen for connections.")
-	debugport        = flag.String("debgport", "localhost:50045", "A separate port for http debug pages.")
+	debugport        = flag.String("debugport", "localhost:50045", "A separate port for http debug pages. Set to an empty string to disable.")
 	credSource       = flag.String("credential-source", mtlsFlags.Name(), fmt.Sprintf("Method used to obtain mTLS creds (one of [%s])", strings.Join(mtls.Loaders(), ",")))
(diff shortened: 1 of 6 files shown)


cmd/proxy-server/main.go
```diff
@@ -64,6 +64,7 @@ var (
 	clientPolicyFlag = flag.String("client-policy", "", "OPA policy for outbound client actions (i.e. connecting to sansshell servers). If empty no policy is applied.")
 	clientPolicyFile = flag.String("client-policy-file", "", "Path to a file with a client OPA.  If empty uses --client-policy")
 	hostport         = flag.String("hostport", "localhost:50043", "Where to listen for connections.")
+	debugport        = flag.String("debgport", "localhost:50045", "A separate port for http debug pages.")
```
sfc-gh-ajoyner, line 67: Please add another sentence like: "Set this to \"\" to disable."

cmd/proxy-server/server/server.go
```diff
@@ -208,6 +244,13 @@ func Run(ctx context.Context, opts ...Option) {
 		}
 	}
 
+	// If there's a debug port, we want to start it early
+	if rs.debughandler != nil {
```
sfc-gh-ajoyner, line 248: I'm not sure != nil allows the caller to disable this with the flag, without building their own main, but adding and != "" would allow them to turn it off with a flag flip.

sfc-gh-ajoyner (approved):
Just minor nits

cmd/proxy-server/main.go
sfc-gh-srhodes: Done.

Your comment also let me notice that I misspelled the flag name :-)

cmd/proxy-server/server/server.go
sfc-gh-srhodes: Sure, done.

PR #194, opened 2023-03-08
Fix typo in build-test.yaml

sfc-gh-srhodes:
I'd misspelled branches, so we were mistakenly building and testing on every push and not only pushes to main


sfc-gh-jallie (approved):
LGTM

PR #195, opened 2023-03-08
Use DialContext instead of Dial in sanssh client

sfc-gh-srhodes:
We have a context available in this function, so we should use it


sfc-gh-jallie (approved):
LGTM

PR #201, opened 2023-03-20
Add Evaluation Result to Authz Audit Logs

sfc-gh-elinardi:
Add authz evaluation result to audit logs to help us understand whether a request is approved or deny.

### DIFF
--- a/auth/opa/rpcauth/rpcauth.go
+++ b/auth/opa/rpcauth/rpcauth.go
@@ -86,10 +86,14 @@ func (g *Authorizer) Eval(ctx context.Context, input *RPCAuthInput) error {
 		}
 	}
 	if input == nil {
-		return status.Error(codes.InvalidArgument, "policy input cannot be nil")
+		errMsg := "policy input cannot be nil"
+		err := status.Error(codes.InvalidArgument, errMsg)
+		logger.V(1).Error(err, errMsg)
+		return err
 	}
 	for _, hook := range g.hooks {
 		if err := hook.Hook(ctx, input); err != nil {
+			logger.V(1).Error(err, "authz hook error", "input", input)
 			if _, ok := status.FromError(err); ok {
 				// error is already an appropriate status.Status
 				return err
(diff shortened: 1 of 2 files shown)


auth/opa/rpcauth/rpcauth.go
```diff
@@ -80,32 +80,38 @@ func NewWithPolicy(ctx context.Context, policy string, authzHooks ...RPCAuthzHoo
...
-	allowed, err := g.policy.Eval(ctx, input)
+	result, err := g.policy.Eval(ctx, input)
 	if err != nil {
+		logger.V(1).Error(err, "failed to evaluate authz policy")
 		return status.Errorf(codes.Internal, "authz policy evaluation error: %v", err)
 	}
-	if !allowed {
-		logger.V(1).Info("permission denied")
+	logger = logger.WithValues("authorizationResult", result)
+	logger.V(1).Info("authz policy evaluation result")
```
sfc-gh-srhodes, line 113: Would it be better to do `.Info("authz policy evaluation result", "authorizationResult", result)`? I only see the logger you're creating being used once

auth/opa/rpcauth/rpcauth.go
```diff
@@ -80,32 +80,38 @@ func NewWithPolicy(ctx context.Context, policy string, authzHooks ...RPCAuthzHoo
...
 			if _, ok := status.FromError(err); ok {
 				// error is already an appropriate status.Status
 				return err
 			}
 			return status.Errorf(codes.Internal, "authz hook error: %v", err)
 		}
 	}
-	if logger.V(1).Enabled() {
-		logger.V(1).Info("evaluating authz policy post hooks", "input", input)
+	if logger.V(2).Enabled() {
```
sfc-gh-srhodes, line 104: Can you remove the if statement?

From looking at some history, it's leftover from when we used to do more work inside the statement prior to https://github.com/Snowflake-Labs/sansshell/pull/151

sfc-gh-srhodes (changes requested):
Nice! My inline comments are minor.

One idea that I haven't fully thought through - should we add a way to let the caller hook into the results to run arbitrary code? I could imagine something like

```
func (g *Authorizer) SetPostEvalHook(func(ctx context.Context, input *RPCAuthInput, allowed bool))
```

auth/opa/rpcauth/rpcauth.go
sfc-gh-elinardi: Done.

auth/opa/rpcauth/rpcauth.go
sfc-gh-elinardi: Agreed. I forgot to remove this after removing a redundant log.

sfc-gh-elinardi:
> Nice! My inline comments are minor.
> 
> One idea that I haven't fully thought through - should we add a way to let the caller hook into the results to run arbitrary code? I could imagine something like
> 
> ```
> func (g *Authorizer) SetPostEvalHook(func(ctx context.Context, input *RPCAuthInput, allowed bool))
> ```

Great idea. It would add more structure to the code and making it easier to follow.
It could be something like:
- PreAuthzHooks()
- Eval()
- PostAuthzHooks()

sfc-gh-srhodes (approved):
I like the idea of an explicit post-eval hook, but that can happen in a different pull request. We should make sure to respect semantic versioning and not break any of the existing public methods and functions when we add them.

PR #203, opened 2023-03-21
Add session id to sansshell logs

sfc-gh-elinardi:
Attach an ID to each sansshell session, and log them.
This would make it easier to trace sansshell logs. We could find all logs generated for a session by looking for its id.


sfc-gh-elinardi:
Replaced by #206

PR #207, opened 2023-03-29
Log authz input after evaluations

sfc-gh-elinardi:
There's a regression resulted from https://github.com/Snowflake-Labs/sansshell/pull/201 that caused authz logs to contain value of `input` before evaluations.
This is because `logger = logger.WithValues("input", input)` is doing copy by value rather than by reference. 
go-logr implementation can be found [here](https://github.com/go-logr/logr/blob/v1.2.3/funcr/funcr.go#L777)

On this PR, we're fixing this issue by adding `input` to every log in `Authorizer.Eval()` so that it always contains the latest value.

Thanks @sfc-gh-srhodes for finding this problem and identifying the root cause quickly.

### DIFF
--- a/auth/opa/rpcauth/rpcauth.go
+++ b/auth/opa/rpcauth/rpcauth.go
@@ -80,18 +80,17 @@ func NewWithPolicy(ctx context.Context, policy string, authzHooks ...RPCAuthzHoo
 // the success or failure of policy.
 func (g *Authorizer) Eval(ctx context.Context, input *RPCAuthInput) error {
 	logger := logr.FromContextOrDiscard(ctx)
-	logger = logger.WithValues("input", input)
 	if input != nil {
-		logger.V(2).Info("evaluating authz policy")
+		logger.V(2).Info("evaluating authz policy", "input", input)
 	}
 	if input == nil {
 		err := status.Error(codes.InvalidArgument, "policy input cannot be nil")
-		logger.V(1).Error(err, "failed to evaluate authz policy")
+		logger.V(1).Error(err, "failed to evaluate authz policy", "input", input)
 		return err
 	}
 	for _, hook := range g.hooks {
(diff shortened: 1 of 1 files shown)


sfc-gh-elinardi:
Verified that Eval logs contain value of `input` after evaluations.
```json
{"logger":"sansshell-proxy","ts":"2023-03-29 12:25:10.665621","caller":{"file":"rpcauth.go","line":107},"level":1,"msg":"authz policy evaluation result","version":"","method":"/Proxy.Proxy/Proxy","peer":{"net":{"network":"tcp","address":"::1","port":"54618"},"cert":{"subject":"SERIALNUMBER=XXXX,O=Acme Co,POSTALCODE={\\\"username\\\":[\\\"edbertlinarditestclient\\\"]\\,\\\"groups\\\":[\\\"APP_Edbert\\\"]}","issuer":"SERIALNUMBER=XXXX,O=Acme Co","dnsnames":["localhost"],"spiffeid":""},"principal":null},"authorizationResult":true,"input":{"method":"/Proxy.Proxy/Proxy","message":{"stream_data":{"stream_ids":["17613228628879050819", "9882643378035907792"], "payload":{"@type":"type.googleapis.com/google.protobuf.Empty", "value":{}}}},"type":"Proxy.ProxyRequest","metadata":{":authority":["localhost:9501"],"content-type":["application/grpc"],"user-agent":["grpc-go/1.53.0"]},"peer":{"net":{"network":"tcp","address":"::1","port":"54618"},"cert":{"subject":"SERIALNUMBER=XXXX","issuer":"SERIALNUMBER=XXXX,O=Acme Co","dnsnames":["localhost"],"spiffeid":""},"principal":{"id":"edbertlinarditestclient","groups":["APP_Edbert","Acme Co"]}},"host":{"net":{"network":"tcp","address":"::","port":"9501"},"cert":null,"principal":null},"environment":null,"extensions":null}}
```

PR #208, opened 2023-03-29
Add unit test for authorization logging

sfc-gh-elinardi:
As a followup to #207, the new tests would make sure that authz logs contain necessary fields, such as peer principal and request method.


sfc-gh-srhodes (approved):
Very nice!

Issue #210, opened 2023-04-03
Symlink handling in `file` action inappropriately follows symlinks


sfc-gh-srhodes:
We're following symlinks when doing actions like listing directories. For example, let's set up a badly-formed directory structure.

```
% mkdir /tmp/repro
% touch /tmp/repro/a /tmp/repro/c  /tmp/repro/d
% ln -s broken /tmp/repro/c
% ln -s /tmp/repro/d /tmp/repro/b
```

Running `ls` normally handles this fine.

```
% ls -l /tmp/repro
total 0
-rw-r--r--@ 1 srhodes  wheel   0 Apr  3 09:11 a
lrwxr-xr-x@ 1 srhodes  wheel  12 Apr  3 09:13 b -> /tmp/repro/d
lrwxr-xr-x@ 1 srhodes  wheel   6 Apr  3 09:12 c -> broken
-rw-r--r--@ 1 srhodes  wheel   0 Apr  3 09:11 d
```

SansShell's file action has issues, following the symlink and failing when it can't do so.

```
% ./sanssh --targets=localhost file ls --long /tmp/repro
-rw-r--r--   -      501        0                0 Mon Apr  3 16:11:57 UTC 2023 /tmp/repro/a
-rw-r--r--   -      501        0                0 Mon Apr  3 16:11:57 UTC 2023 /tmp/repro/b
Got error from target localhost:50042 (0) - rpc error: code = Internal desc = stat: os.Stat error stat /tmp/repro/c: no such file or directory
```

PR #211, opened 2023-04-03
Improve symlink handling.

sfc-gh-srhodes:
Fixes https://github.com/Snowflake-Labs/sansshell/issues/210

I've made dealing with symlinks nicer through adding the following

1. Use lstat when listing files in directory
2. Use lstat by default for `file stat`, imitating the behavior of `stat`
3. Add a new readlink command so that we can read symbolic links
4. Add a new symlink command so that we can create symbolic links
5. Add -l and -d commands to `file ls` because my muscle memory keeps typing them

Some manual testing on macos:

```
% ./sanssh --targets=localhost file ls -l /tmp/repro
-rw-r--r--   -      501        0                0 Mon Apr  3 16:11:57 UTC 2023 /tmp/repro/a
Lrwxr-xr-x   -      501        0               12 Mon Apr  3 16:13:10 UTC 2023 /tmp/repro/b
Lrwxr-xr-x   -      501        0                6 Mon Apr  3 16:12:55 UTC 2023 /tmp/repro/c
-rw-r--r--   -      501        0                0 Mon Apr  3 16:11:57 UTC 2023 /tmp/repro/d
% ./sanssh --targets=localhost file readlink /tmp/repro/b
/tmp/repro/d
% ./sanssh --targets=localhost file readlink /tmp/repro/a
readlink client error: rpc error: code = FailedPrecondition desc = file is not a symlink
```


services/localfile/server/localfile.go
```diff
@@ -628,6 +630,39 @@ func (s *server) Rename(ctx context.Context, req *pb.RenameRequest) (*emptypb.Em
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "readlink error: %v", err)
+	}
+	return &pb.ReadlinkReply{Linkvalue: linkvalue}, nil
+}
+
+func (s *server) Symlink(ctx context.Context, req *pb.SymlinkRequest) (*emptypb.Empty, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+	logger.Info("symlink", "target", req.Target, "linkname", req.Linkname)
+	if err := util.ValidPath(req.Linkname); err != nil {
```
sfc-gh-jallie, line 656: do you also want to check req.Target?

sfc-gh-jallie (approved):
LGTM

services/localfile/server/localfile.go
sfc-gh-srhodes: I skipped the check on purpose. I've added a comment explaining why.

PR #212, opened 2023-04-05
Add a way to expose denial hints in the sansshell policies.

sfc-gh-srhodes:
A common issue when interacting with complex policies is figuring out why a user doesn't have permission to run a command. It's possible to add print statements to diagnose or to grab a copy of the input and run it against a modified local version of a policy, but this tends to be time-consuming and requires knowledge of how policy evaluation works.

This change adds a new way to help with common issues by giving a way to put failure messages into the policy. Normally policies have a single `allow` boolean that gets consulted to decide whether a command is approved. I've added a way to create an additional array of strings that gets queried on a subsequent evaluation after failure.

If you add a line like the following to the rego policy

```
denial_hints [msg] {
  not allow
  msg :="you need to be allowed"
}
```

Then rejected requests will see a failure message like

```
OPA policy does not permit this request: you need to be allowed
```

Enabling this requires passing in an option like `opa.WithDenialHintsQuery("data.sansshell.authz.denial_hints")` when creating the OPA policy.

### DIFF
--- a/auth/opa/opa.go
+++ b/auth/opa/opa.go
@@ -71,7 +71,7 @@ func (o optionFunc) apply(opts *policyOptions) {
 	o(opts)
 }
 
-// WithAllowQuery returns an option to use `query` to evaulate the policy,
+// WithAllowQuery returns an option to use `query` to evaluate the policy,
 // instead of DefaultAuthzQuery. The supplied query should be simple evaluation
 // expressions that creates no binding, and evaluates to 'true' iff the input
 // satisfies the conditions of the policy.
@@ -81,7 +81,7 @@ func WithAllowQuery(query string) Option {
 	})
 }
 
-// WithDenialHintsQuery returns an option to use `query` to evaulate the policy
+// WithDenialHintsQuery returns an option to use `query` to evaluate the policy
 // when the AllowPolicy fails. The supplied query must be a simple evaluation
 // expression that creates no binding and evaluates to an array of strings.
 //
--- a/auth/opa/opa.go
+++ b/auth/opa/opa.go
@@ -22,6 +22,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"sort"
 
 	"github.com/go-logr/logr"
 	"github.com/open-policy-agent/opa/ast"
@@ -49,12 +50,14 @@ var (
 // An AuthzPolicy performs policy checking by evaluating input against
 // a sansshell rego policy file.
 type AuthzPolicy struct {
-	query rego.PreparedEvalQuery
-	b     *bytes.Buffer
+	query            rego.PreparedEvalQuery
+	denialHintsQuery *rego.PreparedEvalQuery
+	b                *bytes.Buffer
 }
 
 type policyOptions struct {
-	query string
+	query            string
+	denialHintsQuery string
 }
 
 // An Option controls the behavior of an AuthzPolicy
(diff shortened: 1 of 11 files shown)


sfc-gh-srhodes:
An example error from my manual sanity check:
```
% ./cmd/sanssh/sanssh --proxy localhost:50043 --targets=localhost file read /etc/host
Target localhost:50042 (0) returned error - rpc error: code = PermissionDenied desc = OPA policy does not permit this request: this message always shows up on errors, we only proxy /etc/hosts
```

auth/opa/opa.go
```diff
@@ -78,6 +81,24 @@ func WithAllowQuery(query string) Option {
 	})
 }
 
+// WithDenialHintsQuery returns an option to use `query` to evaulate the policy
```
sfc-gh-elinardi, line 84: nit: typo `evaluate`

cmd/proxy-server/default-policy.rego
```diff
@@ -53,3 +53,12 @@ allow {
...
 #  input.peer.principal.groups[i] = "admin"
 # }
+
+# Denial reason to help show what went wrong
+denial_hints[msg] {
+	input.message.file.filename != "/etc/hosts"
+	msg := "we only proxy /etc/hosts"
+}
+denial_hints[msg] {
+	msg := "this message always shows up on errors"
```
sfc-gh-elinardi, line 63: is this for testing?

cmd/sansshell-server/main.go
```diff
@@ -127,11 +127,13 @@ func main() {
...
+		log.Fatalf("Invalid policy: %v\n", err)
+	}
+
 	if *validate {
-		_, err := opa.NewAuthzPolicy(ctx, policy)
-		if err != nil {
-			log.Fatalf("Invalid policy: %v\n", err)
-		}
+
 		fmt.Println("Policy passes.")
```
sfc-gh-elinardi, line 137: is this still necessary?

cmd/sansshell-server/default-policy.rego
```diff
@@ -63,3 +63,8 @@ allow {
 allow {
 	input.type = "Service.StatusRequest"
 }
+
+denial_hints[msg] {
+	#input.message.file.filename != "/etc/hosts"
+	msg := "we only allow /etc/hosts"
```
sfc-gh-elinardi, line 69: If I understand this correctly, the denial hint should be placed right after the corresponding rule. So, shouldn't this be at L32?

auth/opa/rpcauth/rpcauth.go
```diff
@@ -106,7 +107,17 @@ func (g *Authorizer) Eval(ctx context.Context, input *RPCAuthInput) error {
 	}
 	logger.V(1).Info("authz policy evaluation result", "authorizationResult", result, "input", input)
```
sfc-gh-elinardi, line 108: Let's log the denial hints. Maybe add an else clause to the if statement, and then add this log to both conditions?
e.g.
```
if result {
  logger.V(1).Info("request approved", "input", input)
} else { // denied
  hints := ...
  logger.V(1).Info("request denied", "input", input, "denialHints", hints)
}
```

sfc-gh-srhodes: I've changed the log to include the hints.

I decided to keep a single log line instead of separate log lines for request approved vs denied. It'll leave an empty denialHints on success, but I think that's okay. I think it's useful to use the same log line for both success and failure.

cmd/sansshell-server/default-policy.rego
sfc-gh-srhodes: One of the weird things about this language is that rule order tends to not matter. Placing after the other rule or at the end both work fine.

https://www.openpolicyagent.org/docs/latest/faq/#statement-order

cmd/sansshell-server/main.go
sfc-gh-srhodes: "run as a one-shot and exit after parsing" was pre-existing, so I left it in to preserve that behavior. We use it in https://github.com/Snowflake-Labs/sansshell/blob/07514978b337858cc74ecf46df814c78610bf4dc/testing/integrate.sh#L466-472 as part of our integration testing.

cmd/proxy-server/default-policy.rego
sfc-gh-srhodes: I put this one on purpose to demonstrate how denial hints work. I've added a comment.

auth/opa/opa.go
sfc-gh-srhodes: Thanks, fixed here and above

sfc-gh-elinardi (approved):

PR #214, opened 2023-04-06
Add gRPC prometheus metrics instrumentations, add new port for metrics endpoint

sfc-gh-elinardi:
We wanted to have better visibility into the proxy server by instrumenting and exporting metrics.
In this PR, I'm separating the `/metrics` endpoint into its own port so that we can get the metrics scraped without enabling the debug port.
Also adding `grpc_prometheus` interceptors which would give us gRPC requests metrics. 
List of the gRPC metrics can be found here: https://github.com/grpc-ecosystem/go-grpc-prometheus#metrics.

# Testing
Ran proxy-server using `WithPrometheusPort`. Was able to fetch the metrics below:
Screenshot 2023-04-05 at 10 30 50 PM

### DIFF
--- a/cmd/proxy-server/server/server.go
+++ b/cmd/proxy-server/server/server.go
@@ -223,6 +223,7 @@ func WithDebugPort(addr string) Option {
 				return
 			}
 		})
+		mux.Handle("/metrics", promhttp.Handler())
 		mux.HandleFunc("/debug/pprof/", pprof.Index)
 		mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
 		mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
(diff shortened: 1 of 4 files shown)


cmd/proxy-server/server/server.go
```diff
@@ -219,7 +223,6 @@ func WithDebugPort(addr string) Option {
 				return
 			}
 		})
-		mux.Handle("/metrics", promhttp.Handler())
```
sfc-gh-srhodes, line 222: I'd like to keep the ability to find raw metrics from the debug port. Could we either keep this registered here or change the index page (line 214) to link to the new port?

I mildly prefer keeping this line and making both the debug port and the metrics port serve /metrics.

cmd/proxy-server/server/server.go
```diff
@@ -231,6 +234,43 @@ func WithDebugPort(addr string) Option {
 	})
 }
 
+func WithPrometheusPort(addr string) Option {
```
sfc-gh-srhodes, line 237: How about we name this WithMetricsPort? The format originally came from Prometheus but it's used by more things now.

Please add a function comment too.

sfc-gh-srhodes (changes requested):

cmd/proxy-server/server/server.go
sfc-gh-elinardi: Agreed. I chose to keep this line so metrics can be accessed at either debug port or metrics port.

cmd/proxy-server/server/server.go
sfc-gh-elinardi: Done

sfc-gh-srhodes (approved):

PR #216, opened 2023-04-12
Add metric instrumentation package, instruments Sansshell-Proxy 

sfc-gh-elinardi:
Adds metric instrumentations to provide visibility into Sansshell-proxy health.

* Create package `telemetry/metrics` for instrumenting metrics. It has a similar behavior to `go-logr` where you create an `MetricsRecorder` object, and add it to the context. Then, you can retrieve the recorder from context as needed.
* Adds gRPC interceptors for passing `MetricsRecorder` between contexts in grpc calls.
* Remove grpc metric instrumentations from `WithMetricsPort()` and adds them to `WithMetricsRecorder()`

List of instrumentations added:
* gRPC server response latency distribution
* authz failures, including OPA policy and justification evaluation failures
* proxy send, and dispatch failures


telemetry/metrics/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	return optionFunc(func(m *Metrics) error {
+		m.prefix = prefix
+		return nil
+	})
+}
+
+// InitMetrics initializes this package by creating the Metrics singleton
+// and applying options.
+// IMPORTANT: you need to call this before adding instrumentations
+func InitMetrics(opts ...Option) error {
```
sfc-gh-srhodes, line 81: I'm not a big fan of doing this particular way of global singletons.

What do you think of instead copying the logr pattern of sticking this in a context and having top-level functions to create a context with it and to pull it out of the context?

```
func NewContext(context.Context, Metrics) context.Context
func FromContext(context.Context) (Metrics, bool)
```

sfc-gh-srhodes (changes requested):
Yay metrics!

telemetry/metrics/metrics.go
sfc-gh-elinardi: I wasn't sure between singleton and context-based, and thought singleton would be simpler.
But, I like the idea of copying the logr pattern because it makes the code more consistent.

sfc-gh-srhodes:
Are there any tests we could add for some of this?

sfc-gh-elinardi:
> Are there any tests we could add for some of this?

This definitely needs some unit tests before it can get merged. Will add it later

sfc-gh-elinardi:
> Are there any tests we could add for some of this?

@sfc-gh-srhodes I added some tests. Could you please take another look? Thank you!

telemetry/metrics/metrics.go
```diff
@@ -0,0 +1,149 @@
...
+// observedStream is a grpc server stream with an attached MetricsRecorder
+type observedStream struct {
+	grpc.ServerStream
+	metricsrecorder MetricsRecorder
+}
+
+// *observedStream.Context returns the server stream context attached with a MetricsRecorder
+func (o *observedStream) Context() context.Context {
+	// Get the stream context and make sure our recorder is attached.
+	ctx := o.ServerStream.Context()
```
sfc-gh-srhodes, line 146: Hmm, I thought that the context used for the stream was derived from our parent context in main(). I guess it isn't? How does logging work if the context isn't getting propogated?

sfc-gh-elinardi: Context used for the grpc stream isn't derived from main. 

Looking at https://github.com/Snowflake-Labs/sansshell/blob/a5c11d0807beb83407eb2172da38948b6e33ae35/cmd/proxy-server/server/server.go#L462 grpc.Serve() doesn't accept context as argument, so context from main gets obsolete as soon as the server starts serving gRPC. 
I briefly looked at the grpc server API. I don't think there's a way to pass context into a grpc server.

We're doing similar thing for logger - interceptors do the job of inserting logger to the context. 
https://github.com/Snowflake-Labs/sansshell/blob/main/telemetry/telemetry.go#L172

cmd/proxy-server/server/server.go
```diff
@@ -263,41 +265,53 @@ func WithDebugPort(addr string) Option {
...
-// This endpoint is to be scraped by a Prometheus-style metrics scraper.
-// It can be accessed at http://{addr}/metrics
-func WithMetricsPort(addr string) Option {
-	return optionFunc(func(_ context.Context, r *runState) error {
-		mux := http.NewServeMux()
-		mux.Handle("/metrics", promhttp.Handler())
-		r.metricsport = addr
-		r.metricshandler = mux
-
+// WithMetricsRecorder enables metric instrumentaitons by inserting grpc metric interceptors
```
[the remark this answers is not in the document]
sfc-gh-elinardi: good catch

telemetry/metrics/metrics.go
sfc-gh-srhodes: Ah, that makes the context solution less clean than I'd hoped. Keeping logging and metrics similar still seems better overall, but just barely.

sfc-gh-srhodes (approved):

PR #220, opened 2023-04-21
Add metric instrumentations to Sansshell services, simplify metrics

sfc-gh-elinardi:
1. Simplify metrics interface by merging `RegisterCounter` and `AddCounter`. I don't see a reason to split them. 
2. Make services emit metrics when they encounter failures. These metrics would help us detect / troubleshoot problems on the machine, or within sansshell itself.


auth/opa/rpcauth/rpcauth.go
```diff
@@ -141,27 +133,20 @@ func (g *Authorizer) Eval(ctx context.Context, input *RPCAuthInput) error {
...
 	if !result {
-		errRegister := recorder.RegisterInt64Counter(authzDeniedPolicyCounterName, authzDeniedPolicyCounterDesc)
+		errRegister := recorder.Counter(ctx, authzDeniedPolicyCounter, 1, attribute.String("method", input.Method))
 		if errRegister != nil {
-			logger.V(1).Error(errRegister, "failed to register "+authzDeniedPolicyCounterName)
-		}
-		errCounter := recorder.AddInt64Counter(ctx, authzDeniedPolicyCounterName, 1, attribute.String("method", input.Method))
-		if errCounter != nil {
-			logger.V(1).Error(errCounter, "failed to add counter "+authzDeniedPolicyCounterName)
+			logger.V(1).Error(errRegister, "failed to add counter "+authzDeniedPolicyCounter.Name)
```
sfc-gh-srhodes, line 149: It seems like we're doing this logging a lot.

Should we have a recorder.CounterOrLog helper function that logs the error instead of returning it?

sfc-gh-elinardi: I like that idea

sfc-gh-srhodes (approved):

PR #238, opened 2023-05-09
Fix `fdbcli fdb kill`

sfc-gh-vrusinov:
Suppose we want to kill an FDB process. One can try to do it naively via sansshell in a following way but it would fail:

```
$ sanssh --targets=${IP?} fdb fdbcli kill 1.2.3.4:4500
fdbcli error: rpc error: code = Internal desc = error from running - <nil>
stdout:

stderr:
ERROR: process `1.2.3.4:4500' not recognized.
```

This is because fdbcli requires kill to be done in two phases: first we get a list of processes by running kill without arguments which fetches list of processes and caches it in the client, so kill <process> works after that. But it’s not convenient to do this via sanshell as it’s not designed for interactive commands. We can do sanssh --targets=${IP?} fdb fdbcli --exec 'kill; kill [proc]' However any kind of --exec reduces our ability to have fine-graned ACLs and auditing.

Sansshell client even documents that `fdbcli kill [proc]` will fetch process list first, however it is not implemented on the server side. This PR fixes it.


sfc-gh-ssudakovich (approved):
LGTM

services/fdb/client/client.go
```diff
@@ -1914,6 +1916,8 @@ NOTE: Internally this will be converted to kill; kill <address...> to do the req
 
 func (r *fdbCLIKillCmd) SetFlags(f *flag.FlagSet) {
 	r.req = &pb.FDBCLIKill{}
+	var d = f.Duration("sleep", time.Duration(0), "Sleep after attempting to kill. Example: -sleep 3s. May be useful to work around lack of https://github.com/apple/foundationdb/pull/9526")
```
sfc-gh-srhodes, line 1919: nit: you can put 0 instead of time.Duration(0) because it's a constant.

services/fdb/fdb.proto
```diff
@@ -276,6 +277,9 @@ message FDBCLIKill {
     FDBCLIKillAll all = 3;
     FDBCLIKillTargets targets = 4;
   }
+  // fdbcli may return before message is actually sent.
```
sfc-gh-srhodes, line 280: Could you add a comment describe what this does? Something like "Duration to sleep after sending the kill command"

services/fdb/server/fdbcli.go
```diff
@@ -899,17 +900,24 @@ func parseFDBCLIKill(req *pb.FDBCLIKill) ([]string, []captureLogs, error) {
 	case *pb.FDBCLIKill_List:
 		args = append(args, "list")
 	case *pb.FDBCLIKill_All:
-		args = append(args, "all")
+		// First kill allready in args will be used cache list of processes.
```
sfc-gh-srhodes, line 903: already, here and below

services/fdb/server/fdbcli.go
```diff
@@ -899,17 +900,24 @@ func parseFDBCLIKill(req *pb.FDBCLIKill) ([]string, []captureLogs, error) {
...
 			return nil, nil, status.Error(codes.InvalidArgument, "kill requires targets")
 		}
+		// First kill allready in args will be used cache list of processes.
+		args = append(args, ";", "kill")
 		args = append(args, t.Addresses...)
 	default:
 		return nil, nil, status.Errorf(codes.InvalidArgument, "unknown request: %T", req.Request)
 	}
 
+	if req.Sleep.AsDuration().Round(time.Second) > 0 {
```
sfc-gh-srhodes, line 917: Style nit: you can avoid repetition if you do

```
if d:= req.Sleep.AsDuration().Round(time.Second); d > 0 {
  		args = append(args, ";", "sleep", strconv.Itoa(int(d.Seconds())))
}
```

sfc-gh-srhodes (approved):

PR #240, opened 2023-05-16
Make service/status handle disabled service

sfc-gh-elinardi:
# Context
Sansshell returns "service not found" when we try to get status of a disabled service. This is due to `ListUnitsContext` not listing disabled services: https://github.com/coreos/go-systemd/blob/v22.5.0/dbus/methods.go#L452.
```
$ ./sanssh --credential-source flags --targets localhost:55001 service status sansshell-server
target localhost:55001 [0] error: rpc error: code = NotFound desc = service sansshell-server was not found
```
Ideally it should be returning something like `status: disabled` instead of not found.

In this PR I'm replacing `ListUnitsContext` with `GetUnitPropertiesContext` which can operate on any services regardless of their status.

Service status can be inferred from UnitProperties, because UnitProperties include `LoadState`, `ActiveState`, and `SubState` which determine the overall service status. 

# Testing
Active service:
```
$ ./sanssh --credential-source flags --targets localhost:55001 service status nginx
[systemd] nginx : running
$ ./sanssh --credential-source flags --targets localhost:55001 service status rsyslog
[systemd] rsyslog : running
```
Disabled service:
```
$ ./sanssh --credential-source flags --targets localhost:55001 service status sansshell-server
[systemd] sansshell-server : stopped
```
Non-existent service:
```
$ ./sanssh --credential-source flags --targets localhost:55001 service status sansshell-servers
[systemd] sansshell-servers : unknown
$ ./sanssh --credential-source flags --targets localhost:55001 service status bob
[systemd] bob : unknown
```

### DIFF
--- a/services/service/server/server_linux.go
+++ b/services/service/server/server_linux.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/coreos/go-systemd/v22/dbus"
+	"github.com/go-logr/logr"
 	"go.opentelemetry.io/otel/attribute"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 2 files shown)


sfc-gh-srhodes:
Is there a minimum systemd version needed for this to work?

sfc-gh-elinardi:
@sfc-gh-srhodes 

For `GetUnitPropertiesContext`, `go-systemd` depends on a systemd method called `org.freedesktop.DBus.Properties.GetAll`: https://github.com/coreos/go-systemd/blob/v22.5.0/dbus/methods.go#L252

Looking at systemd repo, this method has been there since v1 
https://github.com/systemd/systemd/commit/e24067c3ec0e16e94e9620d4f7a06ccc4b637eb4#diff-7f70c17704c959b77a9f27c28c7b8d14b82fc70ebacd79c2df600ad37fa20437R43-R45

sfc-gh-elinardi:
Tested on systemd 219.

sfc-gh-srhodes (approved):

PR #253, opened 2023-06-10
Implement power service

sfc-gh-ssudakovich:
Implement power service

It is useful to be able to reboot a system when necessary. 
To implement that on a linux system, we rely on /sbin/shutdown
command. Other system are not implemented.


services/power/client/client.go
```diff
@@ -0,0 +1,118 @@
+/* Copyright (c) 2019 Snowflake Inc. All rights reserved.
```
sfc-gh-elinardi, line 1: Change year to 2023?

services/power/client/client.go
```diff
@@ -0,0 +1,118 @@
...
+	c := client.SetupSubpackage(subPackage, f)
+	c.Register(&rebootCmd{}, "")
+	return c
+}
+
+func (*powerCmd) Name() string { return subPackage }
+func (p *powerCmd) Synopsis() string {
+	return client.GenerateSynopsis(p.GetSubpackage(flag.NewFlagSet("", flag.ContinueOnError)), 2)
+}
+func (*powerCmd) Usage() string { return "" }
```
sfc-gh-elinardi, line 51: `return client.GenerateUsage(subPackage, p.Synopsis())` might be a better than empty

cmd/sanssh/main.go
```diff
@@ -46,6 +46,7 @@ import (
 	_ "github.com/Snowflake-Labs/sansshell/services/healthcheck/client"
 	_ "github.com/Snowflake-Labs/sansshell/services/localfile/client"
 	_ "github.com/Snowflake-Labs/sansshell/services/packages/client"
+	_ "github.com/Snowflake-Labs/sansshell/services/power/client"
```
sfc-gh-elinardi, line 49: it's not too important, but good idea to also add this to cmd/sansshell-server/main.go

services/power/client/client.go
```diff
@@ -0,0 +1,118 @@
...
+	return `reboot:
+  Reboot a host at a predefined time.
+ `
+}
+func (a *rebootCmd) Usage() string {
+	return client.GenerateUsage(subPackage, a.Synopsis())
+}
+
+func (a *rebootCmd) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&a.when, "when", "now", "When to reboot a host. Valid values are 'now', 'hh:mm', '+5m'.")
```
sfc-gh-elinardi, line 76: `+5m` won't work, so perhaps change this to "`+m` where m is the number of minutes "

services/power/client/client.go
sfc-gh-ssudakovich: Thanks, I did forget what year we live in :-)

services/power/client/client.go
sfc-gh-ssudakovich: Thanks. I should not advertise what is obviously wrong

cmd/sanssh/main.go
sfc-gh-ssudakovich: ack

sfc-gh-elinardi (approved):

PR #257, opened 2023-06-16
add sysinfo service, uptime action

sfc-gh-pchu:
Implement uptime action under sysinfo service

- we can add more system resource checkup action under sysinfo service.
- uptime action is to measure system reliability and stability (how much time the system is up and available to use since last reboot or shutdown).
- the implementation will read /proc/uptime file to get the uptime info since kernel recalculates the value of uptime on the fly and is the most accurate source to deal with uptime.
- uptime action only be supported in linux system right now.


services/sysinfo/server/sysinfo_linux.go
```diff
@@ -0,0 +1,24 @@
...
+	software distributed under the License is distributed on an
+	"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+	KIND, either express or implied.  See the License for the
+	specific language governing permissions and limitations
+	under the License.
+*/
+package server
+
+var getUptimeFilePath = func() (string, error) {
+	return "/proc/uptime", nil
```
sfc-gh-ssudakovich, line 23: Maybe we could use a ready package vs reading it directly. We could use https://pkg.go.dev/golang.org/x/sys/unix.
And then something like
```
var info unix.Sysinfo_t
if err := unix.Sysinfo(&info); err != nil {
    fmt.Println("got error:", err)
}
fmt.Println(time.Duration(info.Uptime) * time.Second)
```

sfc-gh-srhodes: +1

The way to get uptime might vary a lot across operating systems, so implementing a `func() (time.Duration, error)` in both _linux and _default seems better than implementing a `func() (string, error)`.

sfc-gh-pchu: You’re right, in linux we can get the uptime by reading a specific file, but things may not work for other os when we need to support in the future. So providing the `getUptime() `(time.Duration, error) can be more robust. And in linux, both solutions (reading from `/proc/uptime` and` unix.Sysinfo()`) can populate the uptime info well, but the latter solution makes code more concise since it provides a higher-level interface to get system information. So let me change the solution.

cmd/sansshell-server/default-policy.rego
```diff
@@ -68,3 +68,7 @@ denial_hints[msg] {
 	input.message.file.filename != "/etc/hosts"
 	msg := "we only allow /etc/hosts"
 }
+
+allow {
+	input.method = "/SysInfo.SysInfo/Uptime"
+}
```
sfc-gh-srhodes, line 74: nit: add a newline

If you don't want to do this by hand, you can format by installing the opa cli and running `opa fmt -w`

services/sysinfo/client/client.go
```diff
@@ -0,0 +1,105 @@
...
+			retCode = subcommands.ExitFailure
+			continue
+		}
+		d := r.Resp.UptimeSeconds.AsDuration()
+		hours := int(d.Hours())
+		minutes := int(d.Minutes()) % 60
+		seconds := int(d.Seconds()) % 60
+		days := hours / 24
+		hours = hours % 24
+		fmt.Fprintf(state.Out[r.Index], "System_%d (%s) up for %d days, %d hours, %d minutes, %d seconds | total %.f seconds \n", r.Index, r.Target, days, hours, minutes, seconds, d.Seconds())
```
sfc-gh-srhodes, line 102: Is there value in the System_ part of this?

It might be better to be consistent with https://github.com/Snowflake-Labs/sansshell/blob/main/services/healthcheck/client/client.go#L89 and print out a line like

```
10.0.0.1 (0) up for 1 days, 2 hours, 3 minutes, 4 seconds | total 10 seconds
```

cmd/sansshell-server/default-policy.rego
sfc-gh-pchu: Oh, will keep that in mind

services/sysinfo/client/client.go
sfc-gh-pchu: This is what the output looks like:
`$ cd cmd/sanssh && go build && ./sanssh --targets=localhost sysinfo uptime`
`System_0 (localhost:50042) up for 0 days, 19 hours, 34 minutes, 36 seconds | total 70476 seconds `
the number after System_ is the index of the hosts the sanssh client is connecting to.  I think both work fore me, but considering consistency, I will go with your suggestion.

sfc-gh-srhodes (approved):

PR #261, opened 2023-06-23
fix: fdb conf write to the file without changing permission and ownership

sfc-gh-pchu:
Fix the issue that when running `sanssh fdb conf write <section> <key> <value> [--path <config>] command`, the permission and ownership of the file where we modify will change.


sfc-gh-ssudakovich:
For the testing portion, i suspect that the tests and files will be under the same user. I wonder if we could change the user to some other user before the test and then validate that it indeed changed ownership.

sfc-gh-pchu:
Indeed, if the permission and ownership for the temp file is the default one, then there is now way to notice the difference. Let me assign a special permission and ownership for the temp file in advance.

services/fdb/server/conf_test.go
```diff
@@ -86,6 +87,16 @@ cluster_file = /etc/foundatindb/fdb.cluster`)
 	err = f1.Close()
 	testutil.FatalOnErr("can't close tmpfile", err, t)
 
+	// assign a special user, group and permission for this file
+	originUid, originGid, originMod := 1000, 1000, int(0775)
```
sfc-gh-ssudakovich, line 91: You do not know if the userID/GroupID 1000 will exists in the environment you run. GH action would be running in a special environment so there is no guarantee. You can either create new user and chown/grp to it or list existing user and pick a non-root one.

sfc-gh-pchu: And the issue that unit test didn't get passed was because below statement requires root privilege so it threw operation not permitted. And when I ran unit tests in my devVM, they all passed because I was a root user. I found it was hard to grant this operation root like adding sudo in os/exec.Command() which requires password.
`os.Chown(name, originUid, originGid);`
My thought is to mock the chown operation(refer to: https://github.com/Snowflake-Labs/sansshell/blob/main/services/localfile/server/localfile_test.go#L540 ).

It might be the problem that the uid and gid may not be exist, but since I am going to mock the chown operation, not sure whether I still need to pick a real uid and gid.

sfc-gh-ssudakovich: I guess setting it to 1000 is fine. You dont really have to have a user with that ID, it will just not show you a nice username. So it should be fine

sfc-gh-ssudakovich (approved):

PR #263, opened 2023-06-27
WriteRemoteFile supports multiple targets

sfc-gh-elinardi:
# Problem
WriteRemoteFile only supports a single target

# Changes
Enable WriteRemoteFile to accept multiple targets.
- Replace `NewLocalFileClient` with `NewLocalFileClientProxy`, and `Write` to `WriteOneMany` to handle any number of targets
- Send contents in chunks instead of all at once, to handle large files
- Apply the provided Overwrite config

# Testing
- WriteRemoteFile to a target without proxy works
- WriteRemoteFile to a target with proxy works
- WriteRemoteFile to multiple targets with proxy works


services/localfile/client/utils.go
```diff
@@ -68,23 +70,18 @@ type FileConfig struct {
...
 		return fmt.Errorf("can't setup Write stream - %v", err)
 	}
 
-	// Send setup packet
-	if err := stream.Send(&pb.WriteRequest{
+	req := &pb.WriteRequest{
 		Request: &pb.WriteRequest_Description{
 			Description: &pb.FileWrite{
-				Overwrite: true,
+				Overwrite: config.Overwrite,
```
sfc-gh-srhodes, line 84: oh wow, good catch

PR #264, opened 2023-06-29
add fdbserver version lookup

sfc-gh-pchu:
Feature:
- Add the fdbserver service to sanssh
- Currently only version action is supported by fdbserver, no sub-actions

Output looks like:
```shell
$ ./sanssh --targets=localhost fdb fdbserver version  
Target localhost:50042(0) FDB server version:
    FoundationDB 71.2 (v71.2.3)
    source version 2e43b87ddc6df585c960121d9f004b0c9c3ddfae
    protocol fdb00c071200000
```


services/fdb/fdb.proto
```diff
@@ -615,3 +615,18 @@ message FDBCLIResponse {
     Log log = 2;
   }
 }
+
+service Server {
+  rpc FDBServer(FDBServerRequest) returns (FDBServerResponse) {}
+}
+
+message FDBServerRequest {
```
sfc-gh-vrusinov, line 623: This is fine now, but consider how it'll look if we ever need to add something else into `sanssh fdb fdbserver`.

I think these are better as `FDBServerVersionRequest` and `FDBServerVersionResponse`, and the command-line invocation could be `sanssh fdb fdbserver version`

The generic `FDBServerRequest` seems equivalent to `exec run`, so perhaps we don't need it at all?

services/fdb/client/client.go
```diff
@@ -3340,3 +3341,77 @@ func (d *fdbConfDeleteCmd) Execute(ctx context.Context, f *flag.FlagSet, args ..
...
+}
+func (p *fdbServerCmd) Usage() string {
+	return `fdbserver [--version|-v]
+	Run fdbserver for the given flags
+` + client.GenerateUsage(fdbServerPackage, p.Synopsis())
+}
+
+func (r *fdbServerCmd) SetFlags(f *flag.FlagSet) {
+	r.req = &pb.FDBServerRequest{}
+	versionDesc := "Print FoundationDB werver version information and exit."
```
sfc-gh-vrusinov, line 3368:
```suggestion
	versionDesc := "Print FoundationDB server version information and exit."
```

services/fdb/server/fdbserver.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/Snowflake-Labs/sansshell/telemetry/metrics"
+	"go.opentelemetry.io/otel/attribute"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var (
+	// FDBServer string = "/usr/sbin/fdbserver" // uncomment that if fdbserver binary is in that location
+	FDBServer string
```
sfc-gh-vrusinov, line 34: Wait, how does it work? I don't see anywhere where you set it?

Note - if it's just defined in a code it should probably be a constant.

sfc-gh-vrusinov (changes requested):

services/fdb/server/fdbserver.go
sfc-gh-pchu: When I tested the code, just uncommented the line 33 since it defines the fdbserver binary. The reason I uncomment is binary path may be different and to keep consistent with fdbcli: https://github.com/Snowflake-Labs/sansshell/blob/ec9a0e4af5597f44daba1df5ec688b43aecdfd27/services/fdb/server/fdbcli.go#L44 which is set somewhere else.

services/fdb/fdb.proto
sfc-gh-pchu: My original thought was when new actions under fdbserver need to be implemented, we can add the action inside the FDBServerRequest probably name it FDBServerCommand, and the FDBServerCommand will looks like FDBCLICommand: https://github.com/Snowflake-Labs/sansshell/blob/ec9a0e4af5597f44daba1df5ec688b43aecdfd27/services/fdb/fdb.proto#L499 that includes a lot of subcommands. 

Here the version is defined as a flag to fdbserver, I am also fine to put the version as a subaction for fdbserver, in this case the proto will be like :
```go
message FDBServerRequest {
  FDBServerCommand command;
}

message FDBServerCommand {
  oneof command {
     FDBServerVersion version = 1;
  }
}
``` 

Then the action will look like what you suggested `sanssh fdb fdbserver version`.

services/fdb/server/fdbserver.go
sfc-gh-vrusinov: Right, and FDBCLI is set via flag which is commented out here: https://github.com/Snowflake-Labs/sansshell/blob/c4a4183c060a1f9323278c07ae478509a4d79f43/cmd/sansshell-server/main.go#L96

I'm not sure why all FDB stuff is commented out by default, makes it kind annoying. Let't check with people why may know.

sfc-gh-pchu: I will submit a separate PR for uncommenting these lines just in case we may need to revert the PR.

sfc-gh-vrusinov (approved):
LGTM with one comment - feel free to merge after resolving.

You'd need to rebase on top of another PR you merged earlier.

services/fdb/server/fdbserver.go
```diff
@@ -0,0 +1,117 @@
...
+	var args []string
+	for _, c := range req.Commands {
+		newArgs, _, err := parseFDBServerCommand(c)
+		if err != nil {
+			return nil, err
+		}
+		args = append(args, newArgs...)
+		if len(req.Commands) > 1 {
+			args = append(args, ";")
+		}
```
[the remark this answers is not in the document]
sfc-gh-pchu: Yes, since fdbcli we are using --exec to execute all commands, so ; is used to joined different commands. Let me remove it.

PR #266, opened 2023-07-06
Add a sansshell service for proxying HTTP requests.

stvnrhodes:
This adds a structured way to make an HTTP request to a port on localhost. The structure allows fine-grained access control on the http request.

The httpoverrpc proxy subcommand launches a web server that translates http calls into grpc calls, creating a http->grpc->http flow. The webserver only supports one host at a time.

The httpoverrpc get subcommand allows passing in exact arguments to call, giving a way to query the same port across many machines.

A few design notes:
- I've limited the rpc interface to localhost because sansshell is normally used for host-local actions and I don't have a clear use case for making this call cross-host. If we want to make this more general, I could rename things or we could add more RPCs.
- I'm only supporting http calls. Supporting https might require some thought about how to handle certificates and server identity.
- We don't automatically follow redirects because we expect users of the RPCs to do their own redirect-following logic.
- Sufficiently-large http body sizes that are too big to fit into a single grpc message will fail. We could get around size limitations by streaming the response, but that adds complexity to the API.

To test this out, try running the following with each command in a separate terminal.

```
python -m http.Server
go run ./cmd/sansshell-server
go run ./cmd/sanssh -targets localhost httpoverrpc proxy -addr localhost:8001 8000
curl localhost:8001
```


services/httpoverrpc/client/client.go
```diff
@@ -0,0 +1,271 @@
...
+func (*getCmd) Name() string     { return "get" }
+func (*getCmd) Synopsis() string { return "Makes a HTTP call to a port on a remote host" }
+func (*getCmd) Usage() string {
+	return `get [-method METHOD] [-header Header...] [-body body] remoteport request_uri:
+    Make a HTTP request to a specified port on the remote host, defaulting to GET.
+`
+}
+
+func (g *getCmd) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&g.method, "method", "GET", "Address to listen on, defaults to a random localhost port")
```
sfc-gh-jallie, line 206: this comment description seems wrong (accidental paste from 'proxy')

services/httpoverrpc/server/server.go
```diff
@@ -0,0 +1,86 @@
...
+		CheckRedirect: func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse },
+	}
+	httpResp, err := client.Do(httpReq)
+	if err != nil {
+		return nil, err
+	}
+	body, err := io.ReadAll(httpResp.Body)
+	if err != nil {
+		return nil, err
+	}
```
sfc-gh-jallie, line 67: nit: go requires that you always Close() httpResp.Body

services/httpoverrpc/httpoverrpc.proto
```diff
@@ -0,0 +1,54 @@
...
+}
+
+message Header {
+  string key = 1;
+  repeated string values = 2;
+}
+
+// HTTPRequest describes the HTTP request
+message HTTPRequest {
+  string method = 1;
```
sfc-gh-jallie, line 42: for this and 'headers', I wonder about canonicalization and the intersection with policy enforcement. A good policy should naturally not say things like "allow anything where method is not POST", but I could see some confusion since in the sanssh client, "--method" is a flag, so you could pass either --method POST or --method post or --method Post. All of them would "work", but this might make the policy difficult to write.

sfc-gh-jallie (commented):
this is awesome. A few small nits, and one actual question

services/httpoverrpc/httpoverrpc.proto
stvnrhodes: The two options I like:

1. Accept that this can be finicky to use and err towards passing the user's input directly into the RPC call
2. Canonicalize the user's inputs inside the client before passing them into the RPC

Which do you think would be better? I'm leaning a bit towards the implementation simplicity of 1.

services/httpoverrpc/client/client.go
stvnrhodes: Thanks, fixed.

services/httpoverrpc/server/server.go
stvnrhodes: Added

services/httpoverrpc/httpoverrpc.proto
sfc-gh-jallie: 1. is _probably_ fine, if we think we can avoid the policy footguns. 6 months from now, I'm not sure I'd instinctively recognize a policy like:
```
allow { input.method not in ['post', 'delete'] } 
```
as potentially dangerous.

2 probably isn't even the right fix if we're really worried about this. We'd have to canonicalize in the proxy before policy eval, right? (Which we could theoretically do with a authz hook)

stvnrhodes: I've typically seen sansshell policies use allowlists when the set of possible values is small and only switch to blocklists when it's hard to enumerate allowed values. What you mention is a footgun but it doesn't seem much worse than other commands that allow string inputs.

I'd rather not add a precedent of the proxy tampering with input values passed in by the user.

One more option is to switch method to an enum. I think the flexibility of a string is better, but I could be convinced otherwise.

sfc-gh-jallie: I agree that switching to an enum isn't a good option, and wouldn't work for headers, unless we really wanted to limit the functionality.

I don't know that I'd agree that canonicalizing the input before policy eval is "tampering", since that's what the authz hooks are for, in a sense, but I'm ok deferring that discussion for now

sfc-gh-jallie (approved):

PR #267, opened 2023-07-06
Export CmdOptions for unit testing 

sfc-gh-elinardi:
# Problem
It's currently impossible to unit test functions that produce `Option` because of the following reasons:
1. `Option` can't be inspected / compared to another `Option`. The only way to test an `Option` is to apply it to a `cmdOptions` and inspect its values.
2. `apply` method of `Option`, and `cmdOptions` struct is not exported

# Changes
Export `Apply` method, `CmdOptions` and `Env` field, so that `Env` value can be inspected from another package.

### DIFF
--- a/services/util/util.go
+++ b/services/util/util.go
@@ -59,7 +59,7 @@ type CommandRun struct {
 }
 
 // A builder pattern for Options so it's easy to add various ones (such as dropping permissions, etc).
-type cmdOptions struct {
+type CmdOptions struct {
 	failOnStderr bool
 	stdoutMax    uint
 	stderrMax    uint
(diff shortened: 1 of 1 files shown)


sfc-gh-srhodes (changes requested):
I'd rather not expose this public. A big value of the options pattern is that it's opaque to callers and allows easier evolution of the options without breaking backwards compatibility.

Maybe you could add a `OptionsEqual(a, b Option)` function or similar to facilitate testing?

sfc-gh-elinardi:
> I'd rather not expose this public. A big value of the options pattern is that it's opaque to callers and allows easier evolution of the options without breaking backwards compatibility.
> 
> Maybe you could add a `OptionsEqual(a, b Option)` function or similar to facilitate testing?

Makes sense. That's a much better idea!

PR #268, opened 2023-07-06
Add helper function to compare Options

sfc-gh-elinardi:
# Problem
It's currently impossible to unit test functions that produce Option because Option can't be inspected / compared to another Option. The only way to test an Option is to apply it to a cmdOptions and inspect its values.

# Solution
Add helper function `OptionsEqual` and `OptionSlicesEqual` to compare the results of applying two Options.

### DIFF
--- a/services/util/util.go
+++ b/services/util/util.go
@@ -89,7 +89,7 @@ func OptionsEqual(a, b Option) bool {
 	a.apply(aCmdOptions)
 	b.apply(bCmdOptions)
 
-	return cmp.Equal(aCmdOptions, bCmdOptions)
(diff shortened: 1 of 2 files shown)


services/util/util.go
```diff
@@ -80,6 +81,33 @@ func (f optionfunc) apply(opts *cmdOptions) {
 	f(opts)
 }
 
+// OptionsEqual returns true if the results of applying both Options on
```
sfc-gh-srhodes, line 84: Nit: remove the "on an empty cmdOptions" part of the comment. We don't need to reference the unexported object in the doc comments.

sfc-gh-srhodes (approved):

services/util/util.go
sfc-gh-elinardi: Removed. Thanks!

PR #271, opened 2023-07-07
Add packages search by name

sfc-gh-pchu:
Since the packages install & update need **NEVRA**(name-epoch:version-release.architecture) of packages, we provide a way to help user find the NEVRA of package based on the package name:

- --install will fetch the current installed package version
- --available will fetch the available package versions in yum repos
- output will be set to NEVRA
- By default both --installed & --available will be set
- We can set any or both flags of --installed & --available to enable corresponding search mechanism.

This is what the command looks like:
`sanssh packages search --name=firefox [--installed] [--available] [--package_system=P]`
Below is one example output
```shell
$./sanssh --targets=localhost packages search --name firefox --installed --available
Installed Packages:
          firefox-0:102.7.0-1.el7.centos.aarch64
Available Packages:
          firefox-0:78.13.0-2.el7.centos.aarch64
          firefox-0:78.14.0-1.el7.centos.aarch64
          firefox-0:91.2.0-4.el7.centos.aarch64
          firefox-0:91.3.0-1.el7.centos.aarch64
          firefox-0:91.4.0-1.el7.centos.aarch64
          firefox-0:91.6.0-1.el7.centos.aarch64
          firefox-0:91.9.0-1.el7.centos.aarch64
          firefox-0:91.11.0-2.el7.centos.aarch64
          firefox-0:102.5.0-1.el7.centos.aarch64
          firefox-0:102.6.0-1.el7.centos.aarch64
          firefox-0:102.7.0-1.el7.centos.aarch64
          firefox-0:102.8.0-2.el7.centos.aarch64
          firefox-0:102.9.0-3.el7.centos.aarch64
          firefox-0:102.10.0-1.el7.centos.aarch64
          firefox-0:102.11.0-2.el7.centos.aarch64
          firefox-0:102.12.0-1.el7.centos.aarch64

```


services/packages/server/packages.go
```diff
@@ -394,6 +398,24 @@ func parseListInstallOutput(p pb.PackageSystem, r io.Reader) (*pb.ListInstalledR
...
+// extract values from [epoch:]version-release
+func extractEVR(evr string) (epoch, version, release string) {
+	re := regexp.MustCompile(`^([\d]*):(.+)-(.+)$`)
+	matches := re.FindStringSubmatch(evr)
+	if len(matches) == 4 {
+		epoch = matches[1]
+		version = matches[2]
+		release = matches[3]
+	} else {
+		re = regexp.MustCompile(`^(.+)-(.+)$`)
```
sfc-gh-elinardi, line 410: please add comment and examples

sfc-gh-pchu: Yes, will do

services/packages/client/client.go
```diff
@@ -348,8 +349,119 @@ func (l *listCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interfac
...
+	return `search --name=X [--installed] [--latest] [--package_system=P]:
+  Search the current installed packages or the latest packages to be updated on the remote machine. The packages will be displayed in NEVRA version.
+`
+}
+
+func (l *searchCmd) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&l.packageSystem, "package-system", "YUM", fmt.Sprintf("Package system to use(one of: [%s])", strings.Join(shortPackageSystemNames(), ",")))
+	f.StringVar(&l.name, "name", "", "Name of package to search")
+	f.BoolVar(&l.installed, "installed", false, "If true print out installed NEVRA of the package")
+	f.BoolVar(&l.latest, "latest", false, "If true print out latest NEVRA of the package")
```
sfc-gh-elinardi, line 383: I'd set both default to true, otherwise we have to specify a flag on every run.

services/packages/client/client.go
```diff
@@ -348,8 +349,119 @@ func (l *listCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interfac
...
+	// if we don't set flags for current and latest, both should be set by default
+	if !l.installed && !l.latest {
+		installed, latest = true, true
+	}
+	if l.installed {
+		installed = true
+	}
+	if l.latest {
+		latest = true
+	}
```
sfc-gh-elinardi, line 417: I think let's remove these vars and just use the flags

services/packages/packages.proto
```diff
@@ -92,14 +93,40 @@ message UpdateRequest {
...
+message SearchRequest { 
+  PackageSystem package_system = 1; 
+  string name = 2;
+  bool installed = 3;
+  bool latest = 4;
+}
+
+message SearchReply {
+  // key: SearchType value: List of PackageInfo for specific SearchType
+  map<int32, PackageInfoList> package_info_map = 1;
```
sfc-gh-elinardi, line 106: should we do `map<SearchType, PackageInfoList>` instead?

services/packages/server/packages.go
```diff
@@ -161,6 +164,23 @@ var (
...
 
+	generateSearch = func(p *pb.SearchRequest, searchInstalled bool) ([]string, error) {
+		var out []string
+		switch p.PackageSystem {
+		case pb.PackageSystem_PACKAGE_SYSTEM_YUM:
+			out = append(out, RepoqueryBin)
+			out = append(out, "--nevra")
+			out = append(out, p.Name)
+			if searchInstalled {
+				out = append(out, "--installed")
```
sfc-gh-elinardi, line 175: what's the default behavior if there's no `--installed`?

services/packages/packages.proto
sfc-gh-pchu: I am just thinking there will be different search type for this SearchReply, we may want installed or latest or both of package info. In the future we can also expand to get all available package NEVRAs. Map is a way distinguish, or we can also explicitly set fields like:
```go
message SearchReply {
  PackageInfo installed_package = 1;
  PackageInfo latest_package = 2;
}
``` 
The below one is easier to process.

services/packages/server/packages.go
sfc-gh-pchu: Based on https://man7.org/linux/man-pages/man1/repoquery.1.html, repoquery firefox = repoquery -a(-all) firefox  which will display the latest version of that package, but if we set --installed, then the -all wouldn't be set

sfc-gh-pchu: I can add -all to reduce confusion if we want to get latest version

services/packages/client/client.go
sfc-gh-pchu: you are right, I should use the flags

services/packages/client/client.go
sfc-gh-pchu: By default(don't specify any flags), these two flags will be set to true here: https://github.com/Snowflake-Labs/sansshell/pull/271/files#diff-58624590b1dc636ebdcbb5fddf74fa9c9a06d87765bd48593b98e66b2292197dR409 (I will remove the vars and use flags)

services/packages/client/client.go
```diff
@@ -445,7 +544,7 @@ type cleanupCmd struct {
 func (*cleanupCmd) Name() string     { return "cleanup" }
 func (*cleanupCmd) Synopsis() string { return "Run cleanup on machine" }
 func (*cleanupCmd) Usage() string {
-	return `cleanup [--package_system=P]:
+	return `cleanup [--package_system=P]:
```
sfc-gh-vrusinov, line 547: nit: remove trailing spaces

services/packages/packages.proto
```diff
@@ -92,14 +93,34 @@ message UpdateRequest {
 
 message UpdateReply { string debug_output = 1; }
 
-message ListInstalledRequest { PackageSystem package_system = 1; }
+
+message SearchRequest {
```
sfc-gh-vrusinov, line 97: Add some documentation here, e.g. what happens if none of installed & available set, etc.

services/packages/server/packages_test.go
sfc-gh-vrusinov, line 1: I think it would be useful to have a test dedicated just for `extractEVR` - think of all the possible names repoquery may return and ensure it behaves correctly.

sfc-gh-vrusinov (approved):
Approving with a few comments. There are a few iffy things but we've been bike-shedding on it long enough, and current implementation should work well enough.

services/packages/server/packages_test.go
sfc-gh-pchu: I see, let me do it.

sfc-gh-pchu: Actually, the `extractEVR` is used for `yum list` to split its epoch, version and release to make it fit with PackageInfo proto message. The return of repoquery -nevra only has fixed NEVRA format.

PR #273, opened 2023-07-18
Add wait option to sanssh fdb fdbcli exclude action

sfc-gh-pchu:
Sansshell client and server communication right now has 30s timeout, and the fdbcli exclude IP:PORT will move data off the machine which will usually take time over 30s.

- by default sansshell will set **no_wait** for exclude execution. 
- provide **wait** option for user to choose to disable **no_wait**. If set, the command will wait until all data is moved from the machine.
```shell
// by default
$ ./sanssh --targets=localhost fdb fdbcli exclude 127.0.0.1:4502
  127.0.0.1:4502  ---- WARNING: Exclusion in progress! It is not safe to remove this process from the cluster

// set wait to disable no_wait option
$ ./sanssh --targets=localhost fdb fdbcli exclude wait 127.0.0.1:4502
Waiting for state to be removed from all excluded servers. This may take a while.
(Interrupting this wait with CTRL+C will not cancel the data movement.)
  127.0.0.1:4502  ---- Successfully excluded. It is now safe to remove this process from the cluster.
```


services/fdb/client/client.go
```diff
@@ -1361,25 +1365,47 @@ func (r *fdbCLIExcludeCmd) SetFlags(f *flag.FlagSet) {
...
 func (r *fdbCLIExcludeCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interface{}) subcommands.ExitStatus {
 	req := args[1].(*pb.FDBCLIRequest)
+	// by default the no_wait will be set, unless we explicit set the wait option in command
+	r.req.NoWait = &wrapperspb.BoolValue{
+		Value: true,
+	}
 
-	failed, address := false, false
+	wait, failed, address := false, false, false
+	setOpt := func(opt string, optName string, optValue *bool) (bool, subcommands.ExitStatus) {
```
sfc-gh-vrusinov, line 1374: Do we need to return a `bool` too? It seems `subcommands.ExitStatus` can server as an indication of ok/not ok, so  we can return just that.

sfc-gh-vrusinov (approved):

PR #274, opened 2023-07-19
Add /Exec.Exec/StreamingRun

stvnrhodes:
`/Exec.Exec/StreamingRun` is a variant of `/Exec.Exec/Run` that streams back output as the command runs instead of collecting output and returning it at the end. Many commands like ping and tcpdump are commonly used in an interactive way where output is continually printed out and are badly supported by the existing Run call. Streaming the response allows continually printing out output during a call without widening attack surfaces.

StreamingRun uses the exact same request and response as Run. The only difference is that the response is a stream that may be returned many times, each time with a portion of stdout and/or stderr. The final message in the response contains a nonzero exit code if needed.

StreamingRun is a much better experience for anyone using `sanssh exec run`, but we don't want to break backwards compatibility or cause issues when preexisting policies don't allow the new calls or preexisting sansshell nodes haven't been built with the new feature. It can be enabled by passing in `--streaming` to the command or modifying the DefaultStreaming package-level variable in services/exec/client to change the defaults.


stvnrhodes:
@sfc-gh-jallie I want you to check whether this is a feature we're willing to add to sansshell.

@sfc-gh-elinardi can you make sure the implementation seems reasonable?

One quirk of the current cli implementation is that output from different targets gets jumbled together if the data streamed back has no newlines.

services/exec/client/client.go
```diff
@@ -58,21 +65,53 @@ func (p *execCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interfac
...
+
+	// returnCode internally keeps track of the final status to return
+	returnCode subcommands.ExitStatus
+}
 
 func (*runCmd) Name() string     { return "run" }
 func (*runCmd) Synopsis() string { return "Run provided command and return a response." }
 func (*runCmd) Usage() string {
-	return `run <command> [<args>...]:
+	return `run <command> [--streaming] [<args>...]:
```
sfc-gh-jallie, line 78: bikeshed: "--stream" seems equivalently evocative, and is shorter to type (feel free to leave as-is if you disagree)

cmd/sansshell-server/default-policy.rego
```diff
@@ -36,8 +36,8 @@ allow {
 
 allow {
 	input.type = "Exec.ExecRequest"
-	input.message.command = "echo"
-	input.message.args = ["hello", "world"]
+	# input.message.command = "/bin/echo"
```
sfc-gh-jallie, line 39: is this an intentional edit?

sfc-gh-jallie (approved):
LGTM. I think this is a good change

sfc-gh-elinardi (approved):

services/exec/client/client.go
stvnrhodes: I like shorter flag names, changed

cmd/sansshell-server/default-policy.rego
stvnrhodes: Oops, the echo->/bin/echo part was intentional (exec requires absolute filepath) but the rest was leftover from manual testing. Fixed.

PR #275, opened 2023-07-19
Add file mkdir to Sansshell

sfc-gh-pchu:
Implement **sanssh file mkdir** action

Example of running the action:
```shell
$sanssh --targets=localhost file mkdir -username XXX -group XXX -mode 0774 /home/test1
```


services/localfile/localfile.proto
```diff
@@ -267,3 +270,8 @@ message SymlinkRequest {
   string target = 1;
   string linkname = 2;
 }
+
+// MkdirRequest describes the directory
+message MkdirRequest {
+  FileWrite dir_description = 1;
```
sfc-gh-srhodes, line 276: I'd mildly prefer using only FileAttributes

I don't think we should support Overwrite for MkdirRequest. It isn't something supported by `mkdir` and I'm not convinced that there's any behavior we could support for overwrite that isn't surprising.

services/localfile/server/localfile.go
```diff
@@ -774,6 +796,50 @@ func (s *server) Symlink(ctx context.Context, req *pb.SymlinkRequest) (*emptypb.
...
+	}
+
+	cleanup := func() {
+		if retErr != nil {
+			os.Remove(dir)
+		}
+	}
+	defer cleanup()
+
+	// Do one final check (though racy) to see if the file exists.
```
sfc-gh-srhodes, line 829: I don't think we need this check. Directory and file renames are a little different and the directory rename should fail if it tries to get renamed to something preexisting.

services/localfile/server/localfile.go
```diff
@@ -774,6 +796,50 @@ func (s *server) Symlink(ctx context.Context, req *pb.SymlinkRequest) (*emptypb.
...
+	d := req.GetDirDescription()
+
+	a := d.GetAttrs()
+	if a == nil {
+		recorder.CounterOrLog(ctx, localfileMkdirFailureCounter, 1, attribute.String("reason", "missing_attrs"))
+		return nil, status.Errorf(codes.InvalidArgument, "must send attrs")
+	}
+	filename := a.Filename
+	logger.Info("create directory", filename)
+	dir, _, err := setupOutputDir(a)
```
sfc-gh-srhodes, line 816: Looks like you're ignoring the immutable bit. We should either support it or explicitly fail when we see it.

services/localfile/client/client.go
```diff
@@ -1324,3 +1325,114 @@ func (i *renameCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interf
...
+	gid      int
+	group    string
+	mode     int
+}
+
+func (*mkdirCmd) Name() string     { return "mkdir" }
+func (*mkdirCmd) Synopsis() string { return "Create a directory" }
+func (*mkdirCmd) Usage() string {
+	return `mkdir --uid=X|username=Y --gid=X|group=Y --mode=X <path>:
+  Create a directory at the specified path.
```
sfc-gh-srhodes, line 1341: Maybe we should mention not supporting automatically creating intermediate directories? (`mkdir -p`)

sfc-gh-srhodes (commented):
It'd be nice if we could avoid the rename. For files, it's needed to avoid partial writes and to support overwriting an in-use files. For directories, it seems like the chown step is the only thing forcing us to require it to avoid dirs inadvertently owned by root.

What you have seems okay.

services/localfile/server/localfile.go
```diff
@@ -323,6 +325,26 @@ func setupOutput(a *pb.FileAttributes) (*os.File, *immutableState, error) {
 	return f, immutable, err
 }
 
+func setupOutputDir(a *pb.FileAttributes) (string, *immutableState, error) {
```
sfc-gh-vrusinov, line 328: Please add comment/documentation about the purpose of this function

services/localfile/server/localfile.go
```diff
@@ -774,6 +796,50 @@ func (s *server) Symlink(ctx context.Context, req *pb.SymlinkRequest) (*emptypb.
...
+	// Do one final check (though racy) to see if the file exists.
+	_, err = os.Stat(filename)
+	if err == nil {
+		return nil, status.Errorf(codes.Internal, "directory %s already exists ", filename)
+	}
+
+	// Rename tmp directory to real destination.
+	if err := os.Rename(dir, filename); err != nil {
+		return nil, status.Errorf(codes.Internal, "error renaming %s -> %s - %v", dir, filename, err)
+	}
```
sfc-gh-vrusinov, line 838: I'm not clear on the purpose of the create + rename dance for directories. What do we gain by doing it as opposed to just creating directory in the destination right away?

I know there is no atomic way to mkdir+set owner, but the probability of first succeeding and 2nd failing is very low. We are not writing a database here, so maybe we can ignore that in the name of simplicity?

sfc-gh-vrusinov (changes requested):

services/localfile/client/client.go
sfc-gh-pchu: Yes, currently only support creating a directory per action

services/localfile/client/client.go
```diff
@@ -1324,3 +1325,114 @@ func (i *renameCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interf
...
+	f.StringVar(&p.group, "group", "", "The remote file will be set to this group via chown.")
+}
+
+func (p *mkdirCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interface{}) subcommands.ExitStatus {
+	state := args[0].(*util.ExecuteState)
+	if f.NArg() != 1 {
+		fmt.Fprintln(os.Stderr, "Please specify a directory path to create.")
+		return subcommands.ExitUsageError
+	}
+	if (p.uid == -1 && p.username == "") || (p.gid == -1 && p.group == "") || p.mode == -1 {
```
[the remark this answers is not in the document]
sfc-gh-pchu: The first if(p.uid == -1 && p.username == "") ... is to check both uid or username are not set. And latter two if statement (if p.uid >= 0 && p.username != "") and (p.gid >= 0 && p.group != "") are checking whether both the uid and username or both the gid and group are set. Essentially, they are checking different cases.

services/localfile/client/client.go
```diff
@@ -1324,3 +1325,114 @@ func (i *renameCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interf
...
+		fmt.Fprintln(os.Stderr, "cannot set both --uid and --username")
+		return subcommands.ExitFailure
+	}
+	if p.gid >= 0 && p.group != "" {
+		fmt.Fprintln(os.Stderr, "cannot set both --gid and --group")
+		return subcommands.ExitFailure
+	}
+
+	c := pb.NewLocalFileClientProxy(state.Conn)
+	dest := f.Args()[0]
```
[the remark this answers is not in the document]
sfc-gh-pchu: How about directoryName

services/localfile/localfile.proto
sfc-gh-pchu: Oh, FileAttributes is initially the one I want, not sure why picked up FileWrite. Let me change it.

services/localfile/server/localfile.go
sfc-gh-pchu: Yes, I check the documentation: https://pkg.go.dev/os#Rename and does will return error if the directory already exist.

services/localfile/server/localfile.go
sfc-gh-pchu: Yes, that's the main consideration by create a tmp (setting ownership) + rename to make then atomic. But your suggestion also works, we can also directly create a directory and once some wrong happen with ownership or mode setting, os.Remove(dirName). @sfc-gh-srhodes looking forward to your opinion.

services/localfile/server/localfile.go
sfc-gh-pchu: I think adding the immutable to a directory when we create it will leads to the result that no files can be created, renamed or deleted. Then there is not point in creating the directory. So let me explicitly fail when see it.

services/localfile/server/localfile.go
sfc-gh-srhodes: Let's do the simpler in-place implementation for now. We can always change it to a more complex implementation later on if we discover issues.

services/localfile/server/localfile.go
```diff
@@ -323,6 +325,31 @@ func setupOutput(a *pb.FileAttributes) (*os.File, *immutableState, error) {
 	return f, immutable, err
 }
 
+// create a directory, set ownership and mode
+func setupDir(a *pb.FileAttributes) (string, *immutableState, error) {
```
sfc-gh-vrusinov, line 329: maybe `createDir()` or `mkDir()` would be a better name?

sfc-gh-vrusinov (approved):

services/localfile/server/localfile.go
sfc-gh-pchu: I will go with `createDir()`

sfc-gh-srhodes (approved):

PR #277, opened 2023-07-25
Add sysinfo dmesg action to Sansshell

sfc-gh-pchu:
## Description
Allow sansshell to read messages from kernel ring buffer with sysinfo dmesg action. Currently this action is supported only in linux or unix OS. 

## Usage of this action: 
demsg [--tail=N] [--grep=PATTERN] [-i] [-v]:
	 Print the messages from kernel ring buffer.
  -grep string
    	regual expression filter
  -i	ignore case
  -tail int
    	tail the latest n lines (default -1)
  -v	invert match

## Example:
```shell

$./sanssh --targets=localhost sysinfo dmesg -tail 5 -grep USB
[2023-07-24 15:22:10 PDT]: usb 3-2: New USB device strings: Mfr=1, Product=2, SerialNumber=3
 SUBSYSTEM=usb
 DEVICE=c189:264
[2023-07-24 15:22:10 PDT]: hid-generic 0003:203A:FFFB.000C: input,hidraw2: USB HID v10.01 Keyboard [Parallels Virtual Keyboard] on usb-0000:00:03.0-2/input0
 SUBSYSTEM=hid
 DEVICE=+hid:0003:203A:FFFB.000C
[2023-07-24 15:22:10 PDT]: usb 1-1: New USB device found, idVendor=203a, idProduct=fffa, bcdDevice= 1.00
 SUBSYSTEM=usb
 DEVICE=c189:4
[2023-07-24 15:22:10 PDT]: usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
 SUBSYSTEM=usb
 DEVICE=c189:4
[2023-07-24 15:22:10 PDT]: usblp 1-1:1.0: usblp0: USB Unidirectional printer dev 5 if 0 alt 0 proto 1 vid 0x203A pid 0xFFFA
 SUBSYSTEM=usb
 DEVICE=+usb:1-1:1.0
```


services/sysinfo/client/client.go
```diff
@@ -103,3 +105,90 @@ func (p *uptimeCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interf
...
+			fmt.Fprintf(e, "All targets - could not info servers: %v\n", err)
+		}
+		return subcommands.ExitFailure
+	}
+
+	// printed timestamp will be the following format: date time timezone
+	// date: YYYY-MM-DD
+	// time: HH-MM-SS
+	// timezone: MST/PDT etc.
+	timestampFormat := "2006-01-02 15:04:05 MST"
```
sfc-gh-srhodes, line 154: Should we use `time.RFC3339` or another one of the predefined outputs instead of inventing our own?

services/sysinfo/server/sysinfo.go
```diff
@@ -57,6 +62,58 @@ func (s *server) Uptime(ctx context.Context, in *emptypb.Empty) (*pb.UptimeReply
 	return reply, nil
 }
 
+func (s *server) Dmesg(request *pb.DmesgRequest, stream pb.SysInfo_DmesgServer) error {
+	ctx := stream.Context()
+	// logger := logr.FromContextOrDiscard(ctx)
```
sfc-gh-srhodes, line 67: remove commented-out line?

services/sysinfo/server/sysinfo.go
```diff
@@ -57,6 +62,58 @@ func (s *server) Uptime(ctx context.Context, in *emptypb.Empty) (*pb.UptimeReply
...
+	if request.Grep == "" && (request.IgnoreCase || request.InvertMatch) {
+		return status.Error(codes.InvalidArgument, "must provide grep argument before setting ignore_case or invert_match")
+	}
+
+	records, err := getKernelMessages()
+	if err != nil {
+		recorder.CounterOrLog(ctx, sysinfoDmesgFailureCounter, 1, attribute.String("reason", "get kernel message error"))
+		return status.Errorf(codes.InvalidArgument, "can't get kernel message %v", err)
+	}
+	// grep the messages we want
```
sfc-gh-srhodes, line 78: Do you think we could push the grepping into getKernelMessages?

Right now we pull the full contents of kmesg into a slice, then filter down that slice. If we filter the contents as we read them, we avoid holding all the contents we don't care about in-memory.

This likely only makes a difference when there's a lot of kernel messages.

services/sysinfo/server/sysinfo.go
```diff
@@ -57,6 +62,58 @@ func (s *server) Uptime(ctx context.Context, in *emptypb.Empty) (*pb.UptimeReply
...
+		records = newRecords
+	}
+
+	// tail the last n lines
+	// there is no way to tail number of messages more than initial records
+	if request.TailLines > int32(len(records)) {
+		request.TailLines = int32(len(records))
+	}
+	// negative number means disables the tail feature
+	if request.TailLines >= 0 {
```
sfc-gh-srhodes, line 105: I'd minority prefer making this `request.TailLines > 0`. 0 is the default value (https://protobuf.dev/programming-guides/proto3/#default), so returning no data when tail_lines is unset is a little unexpected.

services/sysinfo/server/sysinfo.go
sfc-gh-pchu: Yes have removed in newer commit

services/sysinfo/client/client.go
sfc-gh-pchu: Then I think I will use the default format of time.Time which are also used in file stat action

services/sysinfo/server/sysinfo.go
sfc-gh-pchu: That makes sense to me.

services/sysinfo/server/sysinfo.go
sfc-gh-pchu: It's a good point to reduce memory to hold kernel messages. In this case, the messages in kernel ring buffer has limit. By running below command we can get the size of the buffer which is 1 << 18 = 256 KB
```shell
$ grep CONFIG_LOG_BUF_SHIFT /boot/config-`uname -r`
CONFIG_LOG_BUF_SHIFT=18
``` 
But my initial thought was if we push either grep or tail into the getKernelMessages (function used for a specific OS ), if we want to implement the feature for other OS, the grep and tail will be implemented again but with nearly the same logic.

For journalctl (reading journal log entries), I think we can get only the latest n messages we want.

sfc-gh-srhodes: ehh, if it's <1MB it's probably not worth doing my suggestion. I'm fine with leaving as-is for these messages.

sfc-gh-pchu:
One thing I want to shout out is the timeout setup for kmsgParser: https://github.com/Snowflake-Labs/sansshell/pull/277/files#diff-76044fc118d77e4ad23a112617066226480af94110de9414934321b9460ce8c9R69. Without using that, the [Parse() function](https://github.com/euank/go-kmsg-parser/blob/6575cc091819c8335dec98d8517b11203f5ba93c/kmsgparser/kmsgparser.go#L119) will always keep the channel open which in effect block the program(for loop never ends). 

Although the channel will be closed if it encounters EOF, but it requires us to explicitly call parser.Close() function. The original logic just like this: https://github.com/euank/go-kmsg-parser/blob/master/cmd/godmesg/main.go but the for loop never ends.

So I just wait 2 seconds to close the channel which usually be free within 1 second.

sfc-gh-srhodes:
> One thing I want to shout out is the timeout setup for kmsgParser: https://github.com/Snowflake-Labs/sansshell/pull/277/files#diff-76044fc118d77e4ad23a112617066226480af94110de9414934321b9460ce8c9R69. Without using that, the [Parse() function](https://github.com/euank/go-kmsg-parser/blob/6575cc091819c8335dec98d8517b11203f5ba93c/kmsgparser/kmsgparser.go#L119) will always keep the channel open which in effect block the program(for loop never ends).
> 
> Although the channel will be closed if it encounters EOF, but it requires us to explicitly call parser.Close() function. The original logic just like this: https://github.com/euank/go-kmsg-parser/blob/master/cmd/godmesg/main.go but the for loop never ends.
> 
> So I just wait 2 seconds to close the channel which usually be free within 1 second.

Could you add a comment in the code mentioning this quirk?

It's unfortunate that this library doesn't support a way to only read until we reach the end of recently-logged messages, but I don't think the added complexity from trying to support that is worthwhile yet.

sfc-gh-srhodes (approved):

sfc-gh-pchu:
Yes, actually I have raised a Issue to the owner of the library: https://github.com/euank/go-kmsg-parser/issues/8. And he is working on supporting reading dev/kmsg in non-blocking way. Once the new feature is released, I will make corresponding change on my side.

PR #279, opened 2023-07-28
Add sysinfo journalctl action to Sansshell

sfc-gh-pchu:
## Description
Add journalctl action to Sansshell to read  messages from journald

- input time for since and until should follow the same pattern as "YYYY-MM-DD HH:MM:SS", and in local timezone based on where the sanssh client is run
- the output timestamp is also in local timezone where the sanssh client is run

## Usage
```
journalCtl [--since|--S=X] [--until|-U=X] [-tail=X] [-u|-unit=X] [-o|--output=X] [-x] :
        Get the log entries stored in journald by systemd-journald.service 
  -S string
        Sets the date (YYYY-MM-DD HH:MM:SS) we want to filter from (the date time is included)
  -U string
        Sets the date (YYYY-MM-DD HH:MM:SS) we want to filter until
  -output string
        Sets the format of the journal entries that will be shown. Right now only json and json-pretty are supported.
  -since string
        Sets the date (YYYY-MM-DD HH:MM:SS) we want to filter from
  -tail int
        If positive, the latest n records to fetch. By default, fetch latest 100 records. If negative, fetch all records (default 100)
  -unit string
        Sets systemd unit to filter messages
  -until string
        Sets the date (YYYY-MM-DD HH:MM:SS) we want to filter until (the date time is not included)
  -x    If true, augment log lines with explanatory texts from the message catalog.
``` 

## Example
```shell
./sanssh --targets=localhost sysinfo journalctl -tail 5 -x
[2023-07-28 02:09:00.038149 +0000 UTC]  SDP_DevVM-pchu python3[1504]: Jul 28 02:09:00 1504 INFO] Sleeping
[2023-07-28 02:10:00.086197 +0000 UTC]  SDP_DevVM-pchu python3[1504]: Jul 28 02:10:00 1504 INFO] Sleeping
[2023-07-28 02:10:01.4509 +0000 UTC]  SDP_DevVM-pchu systemd[1]: Started Session 2968 of user root.
-- Subject: Unit session-2968.scope has finished start-up
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- Unit session-2968.scope has finished starting up.
-- 
-- The start-up result is done.
[2023-07-28 02:10:01.45659 +0000 UTC]  SDP_DevVM-pchu CROND[29297]: (root) CMD (/usr/lib64/sa/sa1 1 1)
[2023-07-28 02:11:00.080698 +0000 UTC]  SDP_DevVM-pchu python3[1504]: Jul 28 02:11:00 1504 INFO] Sleeping
```


sfc-gh-srhodes:
Hmm, is this the first C library we're using? What happens if you try to cross-compile across architectures or compile for linux from macos?

sfc-gh-pchu:
Yes, I think it is the first time we are using C library. Crossing compile linux from macos through GOOS=linux will always show this error message:  `pchu@XGX9120JFR sansshell-server % GOOS=linux go build . 
github.com/coreos/go-systemd/v22/internal/dlopen: build constraints exclude all Go files in /XXXXX/go/pkg/mod/github.com/coreos/go-systemd/v22@v22.5.0/internal/dlopen
`

And in devVM(linux env), if we directly compile it, another error will come out: `# github.com/coreos/go-systemd/v22/sdjournal
/Users/pchu/go/pkg/mod/github.com/coreos/go-systemd/v22@v22.5.0/sdjournal/journal.go:27:11: fatal error: 'systemd/sd-journal.h' file not found
 #include <systemd/sd-journal.h>
          ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.`
It is easy to reproduce, here is the previous merge-gate failure: https://github.com/Snowflake-Labs/sansshell/actions/runs/5694246008.

The **reason** is that Go compiler is unable to find the systemd/sd-journal.h file while trying to compile the github.com/coreos/go-systemd/v22/sdjournal package. This file is part of the systemd library, which provides an API for interacting with the systemd journal. 

The **resolution** is to install the development package for systemd utility library like systemd-devel or  libsystemd-dev. 
Then building the project in linux system(container) won't have any issue.

But I don't have a chance to cross-compile across architecture, let me test it.

sfc-gh-pchu:
I understand it increases the complexity in compiling sansshell when introducing C package. I import sdjournal for following reasons:

- it is the most popular [golang systemd packages](https://github.com/coreos/go-systemd)
- the [data fields](https://github.com/coreos/go-systemd/blob/d843340ab4bd3815fda02e648f9b09ae2dc722a7/sdjournal/journal.go#L337) it has  can greatly fit into our use case and easy to extend.

But let me think other workaround.

sfc-gh-elinardi:
Closing this in favor of https://github.com/Snowflake-Labs/sansshell/pull/293

PR #288, opened 2023-08-04
Change directory for temporary files

sfc-gh-srhodes:
atomicSaveTo is implemented via writing a file to a temporary directory, then renaming the file. Renaming is an atomic operation on linux as long as the source and destination are on the same filesystem partition.

The current code leaves the directory unspecified, which defaults to /tmp. If /tmp is on a different partition (like tmpfs or a bind-mount), the later rename will fail. Using the same directory as the final file should be much more likely to succeed.

### DIFF
--- a/services/fdb/server/conf.go
+++ b/services/fdb/server/conf.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"io/fs"
 	"os"
+	"path/filepath"
 	"syscall"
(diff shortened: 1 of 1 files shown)


sfc-gh-hsalih (approved):

sfc-gh-ssudakovich (approved):
LGTM

PR #290, opened 2023-08-04
Update sansshell httpproxy to make it have ability to send http/https…

sfc-gh-pchu:
## Description

- allow sanssell proxy http/https request to remote host
- leave the original proxy action as it is 
- make changes to get action and add protocol and hostname flag

## usage
Note: how the request reply rely on the implementation of the endpoint
```shell
./sanssh --targets=localhost httpoverrpc get -body '{"id": "XXXXX", "price":10, }' -header 'content-type: application/json' -method POST -protocol https  -hostname localhost  443 /place-order
{"code":null,"message":"place order successfully","success":true}
```


services/httpoverrpc/client/client.go
```diff
@@ -136,16 +136,18 @@ func (p *proxyCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interfa
...
 			Request: &pb.HTTPRequest{
 				RequestUri: httpReq.RequestURI,
 				Method:     httpReq.Method,
 				Headers:    reqHeaders,
 				Body:       body,
 			},
-			Port: int32(port),
+			Port:     int32(port),
+			Protocol: "http",
+			Hostname: "localhost",
```
sfc-gh-elinardi, line 148: should this be `httpReq.Protocol` and `httpReq.Hostname`?

services/httpoverrpc/client/client.go
```diff
@@ -136,16 +136,18 @@ func (p *proxyCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interfa
...
-		req := &pb.LocalhostHTTPRequest{
+		req := &pb.HostHTTPRequest{
 			Request: &pb.HTTPRequest{
 				RequestUri: httpReq.RequestURI,
 				Method:     httpReq.Method,
 				Headers:    reqHeaders,
 				Body:       body,
 			},
-			Port: int32(port),
+			Port:     int32(port),
```
sfc-gh-elinardi, line 146: let's add validation for port number

services/httpoverrpc/client/client.go
```diff
@@ -192,18 +194,22 @@ type getCmd struct {
...
-    Make a HTTP request to a specified port on the remote host, defaulting to GET.
+	return `get [-method METHOD] [-header Header...] [-body body] [-protocol Protocol] [-hostname Hostname] remoteport request_uri:
+    Make a HTTP request to a specified port on the remote host.
 `
 }
 
 func (g *getCmd) SetFlags(f *flag.FlagSet) {
 	f.StringVar(&g.method, "method", "GET", "Method to use in the HTTP request")
+	f.StringVar(&g.protocol, "protocol", "http", "protocol to communicate with specified hostname")
+	f.StringVar(&g.hostname, "hostname", "localhost", "ip address or domain name to specify host")
```
sfc-gh-elinardi, line 212: With https://github.com/Snowflake-Labs/sansshell/pull/283, we need to enable `--allow-any-host` flag, so let's add a note here

services/httpoverrpc/client/client.go
sfc-gh-pchu: Oh you mean we should put protocol and hostname inside HTTPRequest rather than HostHTTPRequest definition. To be honest, I think we can put all fields in HostHTTPRequest to HTTPRequest like (port, hostname, protocol), since they are all parts of the http request.

services/httpoverrpc/client/client.go
sfc-gh-pchu: Yes, I will verify the port number fall into the range of [0~65535]

services/httpoverrpc/client/client.go
sfc-gh-pchu: Or do you mean we should use http.Request.Host rather than a string to represent the hostname?

sfc-gh-elinardi: My bad I misunderstood. I think we can keep this as it is.

sfc-gh-elinardi (approved):

services/httpoverrpc/httpoverrpc.proto
```diff
@@ -22,14 +22,18 @@ package HTTPOverRPC;
 
 // The HTTPOverRPC service definition
 service HTTPOverRPC {
-  // Make an HTTP call to localhost
-  rpc Localhost(LocalhostHTTPRequest) returns (HTTPReply) {}
+  // Make an HTTP call to specified host
+  rpc Host(HostHTTPRequest) returns (HTTPReply) {}
```
sfc-gh-srhodes, line 26: This change is slightly dangerous. Renaming the RPC method name in-place breaks backwards compatibility because the method name's passed as a string over the grpc connection to decide which method to call.

However, I think it's okay because the command's been around for less than a month (https://github.com/Snowflake-Labs/sansshell/pull/266) and it's unlikely that anyone's picked up a strong dependency on that at this time. Previous behavior can be achieved on a policy level with

```
allow {
  input.message.name == "localhost"
  input.message.protocol == "http"
}
```

cc @sfc-gh-jallie because he helped review the original PR.

sfc-gh-srhodes (approved):

services/httpoverrpc/httpoverrpc.proto
sfc-gh-pchu: I made the change in place also because this action was release very recently, but the best way is to create a new rpc method and deprecate the old one gradually, right? So here should I keep the Localhost rpc method or continue with the current approach.

sfc-gh-srhodes: I'm happy with the current approach.

If the feature were around for more than 6-12 months, we could mark it as deprecated but we'd want to keep it around indefinitely for backwards compatibility.

PR #292, opened 2023-08-07
Add a way to set FDB-related env vars from a file.

sfc-gh-srhodes:
This is conceptually similar to FDBCLIEnvList, but it's based on reading from a file instead of reading from environmental variables from the sansshell process. Sansshell often runs in a way that has minimal environmental variables, so foundationdb-specific variables may not be around. We silently omit the file when it's absent.


sfc-gh-sgwydir (approved):
LGTM!

sfc-gh-ssudakovich (approved):
LGTM

PR #294, opened 2023-08-07
Add idle timeout and dial timeout

sfc-gh-elinardi:
This PR changes sanssh timeout behaviors. 
1. Dial Timeout
  * Currently dial timeout can only be specified in the targets or proxy address, such as `--proxy=localhost:50043;5s`. If it's not set, we'll use the default grpc dial timeout which is 20s.
  * I'm adding a new flag `--dial-timeout`, default to 10s. I'm hoping that this is easier to use compared to specifying timeout in addresses. 
  This will be the dial timeout used if the target or proxy address doesn't contain a timeout. In other words, timeout specified in the target or proxy address takes precedence over this flag.
  
2. Idle Timeout
  * Currently, `--timeout` flag (default to 30s) is used to define the duration to wait for the command to complete. The timeout counter doesn't reset even if there are responses from the server, which is problematic for streaming RPCs because we don't want to terminate the command while server is still progressing.
  * I'm renaming this to `--idle-timeout`, default to 15m. Now the timeout will reset when there's a response from the server. It will timeout when there's no new responses for the duration specified in the flag. This is done by adding gRPC interceptors that add deadline to every Send or Recv call. 

Rationale for default timeout values (10s dial timeout, and 15m idle timeout): we want to fail fast during initial connection, but it's okay to wait a bit longer for a target that's initially responsive.

### DIFF
--- a/cmd/sanssh/main.go
+++ b/cmd/sanssh/main.go
@@ -62,14 +62,16 @@ const (
 )
 
 var (
-	defaultTimeout = 30 * time.Second
+	defaultDialTimeout = 30 * time.Second
+	defaultIdleTimeout = 15 * time.Minute
 
 	proxyAddr = flag.String("proxy", "", fmt.Sprintf(
 		`Address (host[:port]) to contact for proxy to sansshell-server.
 %s in the environment can also be set instead of setting this flag. The flag will take precedence.
 If blank a direct connection to the first entry in --targets will be made.
 If port is blank the default of %d will be used`, proxyEnv, defaultProxyPort))
-	timeout          = flag.Duration("timeout", defaultTimeout, "How long to wait for the command to complete")
+	dialTimeout      = flag.Duration("timeout", defaultDialTimeout, "How long to wait for the connection to be accepted")
+	idleTimeout      = flag.Duration("idle-timeout", defaultIdleTimeout, "Maximum time that a connection is idle. If no messages are received within this timeframe, connection will be terminated")
 	credSource       = flag.String("credential-source", mtlsFlags.Name(), fmt.Sprintf("Method used to obtain mTLS credentials (one of [%s])", strings.Join(mtls.Loaders(), ",")))
 	outputsDir       = flag.String("output-dir", "", "If set defines a directory to emit output/errors from commands. Files will be generated based on target as destination/0 destination/0.error, etc.")
 	justification    = flag.String("justification", "", "If non-empty will add the key '"+rpcauth.ReqJustKey+"' to the outgoing context Metadata to be passed along to the server for possible validation and logging.")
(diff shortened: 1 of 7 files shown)


cmd/sanssh/main.go
```diff
@@ -62,14 +62,18 @@ const (
...
+	defaultIdleTimeout = 15 * time.Minute
 
 	proxyAddr = flag.String("proxy", "", fmt.Sprintf(
 		`Address (host[:port]) to contact for proxy to sansshell-server.
 %s in the environment can also be set instead of setting this flag. The flag will take precedence.
 If blank a direct connection to the first entry in --targets will be made.
 If port is blank the default of %d will be used`, proxyEnv, defaultProxyPort))
-	timeout          = flag.Duration("timeout", defaultTimeout, "How long to wait for the command to complete")
+	// Deprecated: --timeout flag is deprecated. Use --idle-timeout or --dial-timeout instead
+	_                = flag.Duration("timeout", defaultDialTimeout, "DEPRECATED. Please use --idle-timeout or --dial-timeout instead")
```
sfc-gh-srhodes, line 74: Right now this is silently ignored.

I think we should remove it entirely. This main function is an example of how users of sansshell should implement sanssh, so we can get away with breaking changes.

The other option is sending a message to stderr when this is set and possibly exiting with an error code.

cmd/util/util.go
```diff
@@ -60,14 +61,16 @@ func hasPort(s string) bool {
 	return strings.LastIndex(s, "]") < strings.LastIndex(s, ":")
 }
 
-// ValidateAndAddPort will take a given target address and optionally add a default port
-// onto it if a port is missing. This will also take in account optional dial timeout
-// suffix and make sure the returned value has that if it exists.
-func ValidateAndAddPort(s string, port int) string {
```
sfc-gh-srhodes, line 66: This is part of our public API. Can we add the new function without removing the old one?

Maybe you can make this

func ValidateAndAddPort(s string, port int) string {
  return ValidateAndAddPortAndTimeout(s, port, 0)
}

testing/integrate.sh
```diff
@@ -628,16 +627,16 @@ fi
 mkdir -p ${LOGS}/exec-testdir
 ${SANSSH_PROXY} ${MULTI_TARGETS} --output-dir=${LOGS}/exec-testdir exec run /bin/sh -c "echo foo >&2"
 check_status $? /dev/null exec failed emitting to stderr
-if [ ! -f ${LOGS}/exec-testdir/0-localhost:50042 ] || [ ! -f ${LOGS}/exec-testdir/1-localhost:50042 ]; then
+if [ ! -f "${LOGS}/exec-testdir/0-localhost:50042;10s" ] || [ ! -f "${LOGS}/exec-testdir/1-localhost:50042;10s" ]; then
```
sfc-gh-srhodes, line 630: I suspect creating files with a filename containing `;` will do more harm than good. Can we strip the duration out of the filename?

sfc-gh-srhodes (changes requested):
I like this improvement.

cmd/sanssh/main.go
sfc-gh-elinardi: Updated to run log.Fatalf when this is set

cmd/util/util.go
sfc-gh-elinardi: Thanks for catching this! Added it back

testing/integrate.sh
sfc-gh-elinardi: Good point.

sfc-gh-srhodes (approved):

PR #300, opened 2023-08-28
ReadRemoteFile supports multiple targets

sfc-gh-elinardi:
Enable ReadRemoteFile to accept multiple targets.
- NewLocalFileClientProxy instead of NewLocalFileClient
- ReadOneMany instead of Read
- Create a new struct which includes [Target, Index, Contents] for returning results from multiple targets

Also updating the expired certs in mtls/testdata.


services/localfile/client/utils.go
```diff
@@ -13,28 +13,31 @@ import (
...
-	}
+type ReadRemoteFileResponse struct {
+	Target   string
+	Index    int
+	Contents []byte
+}
 
-	c := pb.NewLocalFileClient(conn)
-	stream, err := c.Read(ctx, &pb.ReadActionRequest{
+// ReadRemoteFile is a helper function for reading a single file from a remote host using a proxy.Conn.
```
sfc-gh-srhodes, line 22: Update the comment to be "from one or more hosts" ?

sfc-gh-srhodes, line 22: This is part of our public API (even though it's buried a bit deeply), so I'd like to avoid unnecessarily changing the function signature in backwards compatible ways. What do you think of adding a new function like ReadRemoteFileFromMany and making ReadRemoteFile a small wrapper around it?

sfc-gh-elinardi: Great point! I'll take that suggestion

sfc-gh-elinardi: Done

sfc-gh-srhodes (approved):

PR #297, opened 2023-08-11
Add generic knob parsing for fdbcli

sfc-gh-jfu:
Sample output:
>`[jfu@awsuswest2temptest000032 ~]$ ./sanssh --targets localhost:50042 fdb fdbcli --tls_certificate_file /etc/foundationdb/fdb.pem --tls_ca_file /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem --tls_key_file /etc/foundationdb/fdb.pem getversion`

169177500851515

> `[jfu@awsuswest2temptest000032 ~]$ ./sanssh --targets localhost:50042 fdb fdbcli --tls_certificate_file /etc/foundationdb/fdb.pem --tls_ca_file /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem --tls_key_file /etc/foundationdb/fdb.pem  --knobs key_size_limit=100,value_size_limit=100 getversion`

1691774576431311

> `[jfu@awsuswest2temptest000032 ~]$ ./sanssh --targets localhost:50042 fdb fdbcli --tls_certificate_file /etc/foundationdb/fdb.pem --tls_ca_file /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem --tls_key_file /etc/foundationdb/fdb.pem  --knobs key_size_limit=100,value_size_limit=100,ab getversion`

fdbcli error: rpc error: code = Internal desc = can't parse knob: ab

> `[jfu@awsuswest2temptest000032 ~]$ ./sanssh --targets localhost:50042 fdb fdbcli --tls_certificate_file /etc/foundationdb/fdb.pem --tls_ca_file /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem --tls_key_file /etc/foundationdb/fdb.pem  --knobs key_size_limit=100,value_size_limit=100,ab=2 getversion`

1691774584442933
WARNING: Invalid knob option 'ab'

> `[jfu@awsuswest2temptest000032 ~]$ ./sanssh --targets localhost:50042 fdb fdbcli --tls_certificate_file /etc/foundationdb/fdb.pem --tls_ca_file /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem --tls_key_file /etc/foundationdb/fdb.pem  --knobs key_size_limit=100,value_size_limit=100c getversion`

1691774591720497
WARNING: Invalid value '100c' for knob option 'value_size_limit'

[...]

PR #327, opened 2023-09-21
Restore old proto tags

sfc-gh-srhodes:
In https://github.com/Snowflake-Labs/sansshell/pull/297, we changed a proto tag number from 19 to 20, breaking backwards compatibility. This restores the old tag number.

I'm reserving tag 20 because I prefer never reusing fields that were used before, even briefly.


sfc-gh-ldormoy (approved):
LGTM

Issue #346, opened 2023-10-11
Multi Party Authorization


sfc-gh-srhodes:
Within Snowflake, we're encountering some use cases where we'd benefit from the ability to require one or more human approvals to a Sansshell command at a per-command granularity. This issue tracks any changes related to adding that feature.

PR #350, opened 2023-10-13
Add a MPA design and proto definition.

stvnrhodes:
I've prototyped this enough in https://github.com/stvnrhodes/sansshell/tree/mpa to have reasonable confidence that this design is implementable. I'm separating out the design and proto into an initial PR so that we can make sure that we agree on the high-level direction before we deeply review any implementation.

The readme is meant to be used as long-term documentation on how to use MPA.


services/mpa/README.md
```diff
@@ -0,0 +1,192 @@
...
+   justification: emergency
+   method: /Exec.Exec/Run
+   message: {
+      "command": "/bin/echo",
+      "args": ["hi"]
+   }
+   $ sanssh -targets=1.2.3.4 mpa approve 1-2345-6789
+   ```
+
+3. If the user's command is still running, it will complete. If the user had stopped their command, they can rerun it and the approval will still be valid as long as the command's input remains the same.
```
sfc-gh-ajoyner, line 35: nit: should probably say “as long as the command’s input remains the same and the sansshell-server approval is still in memory (the binary has not restarted, and it has not been cleared by a user)"

services/mpa/mpa.proto
```diff
@@ -0,0 +1,120 @@
...
+
+message Principal {
+  // The principal identifier (e.g. a username or service role)
+  string id = 1;
+  // Auxiliary groups associated with this principal.
+  repeated string groups = 2;
+}
+
+
+message StoreRequest {
```
sfc-gh-ajoyner, line 74: It might be helpful for StoreRequest (and StoreResponse) to have a TTL.  Eventually you'll want to expire requests that weren't consumed, or it'll OOM sansshell-server in the limit conditions.  A client default of 72hrs or something probably puts an upper bound on the memory usage, without unnecessarily hampering typical usage.  WDYT?

sfc-gh-ajoyner: Is there also a plan to cap the number of in-flight MPA requests that sansshell-server is willing to queue?  That would be another defense that might make the TTL unnecessary.

services/mpa/mpa.proto
```diff
@@ -0,0 +1,120 @@
...
+}
+
+message GetRequest { string id = 1; }
+
+message GetResponse {
+  Action action = 1;
+  
+  // All approvers of the request.
+  repeated Principal approver = 2;
+}
```
sfc-gh-ajoyner, line 117: This will sometimes contain sensitive data.  The (reasonable) advice from the README.md "It's highly recommended to additionally allow `/Mpa.Mpa/List`, `/Mpa.Mpa/Get`" runs the risk that you'll expose that sensitive data to a wide audience.  Complex policies could be written, but perhaps a simpler approach is for `StoreRequest` to allow the user to send a read ACL, which `Get` could enforce?  WDYT?

sfc-gh-ajoyner (approved):
The big picture looks right, so I'm going to go ahead and smash Approve.  I've left some comments we can discuss here, or that you can consider food-for-thought as the implementation develops.  None of them are show-stoppers, for sure.

services/mpa/mpa.proto
stvnrhodes: I have a "Requests are stored in-memory and older requests may be cleared automatically" note in the RPC method comment that' should cover this.

I did "evict oldest request if there's >1000" in my prototype.  It's probably worth adding an "invalidate after X hours" so that old approvals don't cause confusion later on.

services/mpa/mpa.proto
stvnrhodes: This part's a bit tricky. Get intentionally has nothing other than an ID in the request so rego policies can't do useful access control based on the request contents. I don't want to stick a read acl in StoreRequest for a couple reasons.

1. Users would have to specify more than `-mpa` when creating an MPA request
2. The access control is different than anything else we have and we'd either have a non-rego way of doing it or we'd have to have a weird rpcauth hook to allow doing it through rego

I've tweaked the wording a bit to say "let approvers do it" instead of "let anyone do it". In practice I think matching against a combination of peer identity and host information can give reasonable protection on who sees the data..

services/mpa/README.md
stvnrhodes: Good call, updated.

PR #344, opened 2023-10-09
Use go 1.21

sfc-gh-srhodes:
This allows us to use context.WithoutCancel and there's no expected incompatible breaking changes.

PR #357, opened 2023-10-19
Revert "Use go 1.21 (#344)"

sfc-gh-elinardi:
This reverts commit 04285e2eba7489ef231fedc1e98ab98964efa726.
Downgrading to keep the minimum go version for importing sansshell at 1.19.


sfc-gh-srhodes (approved):
Could you add a bit more details to the PR description on why we're doing this?

sfc-gh-elinardi:
> Could you add a bit more details to the PR description on why we're doing this?

Added!

PR #358, opened 2023-10-20
Add interceptors to the proxy connection

stvnrhodes:
This is heavily based on gRPC interceptors from https://github.com/grpc/grpc-go/blob/master/examples/features/interceptor/README.md and serve the same purpose for proxy-level interactions. These interceptors allow hooking into calls before they get fanned out to many targets. The public interface is new fields on the `proxy.Conn` struct because I'm introducing these in a backwards-compatible way.

I'm planning on using this for MPA requests in sanssh. The interceptor lets us hook into every call and turn it into a series of calls that perform the MPA request flow and then pass on the MPA request as metadata into the call. I'm not completely happy with separate calls for unary and streaming, especially because a message with a single request and a streamed response counts as "streaming", but it matches what gRPC does in its calls.

It's theoretically possible to avoid adding this interceptor and instead do what I'd like via a grpc.StreamClientInterceptor. Doing so would require writing something tightly coupled to the logic in the proxy client code and more prone to bugs. It would involve weird things like lying to the client and claiming that we've connected to targets before we connect to any because the message containing the rpc request doesn't get sent until the proxy client code thinks it has established all connections.


sfc-gh-elinardi (approved):
neat!

PR #359, opened 2023-10-20
Pass over caller identity information through the sansshell proxy

stvnrhodes:
For multi party authentication support, we plan on storing state in-memory in the sanshell server so the server needs to know the original caller's identity. Right now that gets lost when the caller goes through the proxy.

This PR fixes that by including a json blob of marshalled rpcauth.PrincipalAuthInput information in the gRPC context sent from the proxy to the server. We use json and not base64-encoded proto because the size is unlikely to be significant and there's no existing common proto with the fields we want. I'm unconditionally sending the information when principal is populated so that we don't expose an excessive number of behavioral knobs.

The API design intentionally tries to avoid exposing easy ways to use the gRPC metadata without checking its validity. We assume that in some cases sansshell server will have both proxied clients and direct clients. Most direct clients should be unable to pass in arbitrary proxied identity information.

I've updated the MPA service to make use of the proxied information. It's currently the only user.


proxy/server/server.go
```diff
@@ -274,6 +274,10 @@ func dispatch(ctx context.Context, requestChan chan *pb.ProxyRequest, replyChan
 				streamSet.ClientCloseAll()
 				return nil
 			}
+			// Peer information might not be properly populated until rpcauth
+			// evaluates the initial received message, so let's grab fresh
+			// peer information when we know we've gotten at least one message.
+			ctx = rpcauth.AddPeerToContext(ctx, rpcauth.PeerInputFromContext(stream.Context()))
```
sfc-gh-jallie, line 280: I have a vague worry about this generating many, many contexts. Walking the codepath here, AddPeerToContext calls WithContext, which creates a new context object on each call. It seems 'cheap' to add a boolean here to track whether we've done this dance once, and avoid doing it every time we receive a message on the channel

auth/opa/proxiedidentity/proxiedidentity.go
```diff
@@ -0,0 +1,137 @@
...
+type proxiedIdentityKey struct{}
+
+// newContext creates a new context with the identity attached.
+func newContext(ctx context.Context, p *rpcauth.PrincipalAuthInput) context.Context {
+	return context.WithValue(ctx, proxiedIdentityKey{}, p)
+}
+
+// FromContext returns the identity in ctx if it exists. It will typically
+// only exist if ServerProxiedIdentityUnaryInterceptor was used.
+func FromContext(ctx context.Context) (p *rpcauth.PrincipalAuthInput, ok bool) {
```
sfc-gh-jallie, line 65: I kinda hate named return values, and I don't think they add to the readability here. (generally, the only time I find them useful is when you have multiple return values of identical type)

auth/opa/proxiedidentity/proxiedidentity.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// We don't expose the key so that people aren't tempted to use it directly.
+// We intentionally prefix this with something other than sansshell- so that
+// https://github.com/Snowflake-Labs/sansshell/blob/main/telemetry/telemetry.go
+// doesn't accidentally send untrusted data in passAlongMetadata().
+const reqProxiedIdentityKey = "proxied-sansshell-identity"
+
+// ServerProxiedIdentityUnaryInterceptor adds information about a proxied caller to the RPC context
+// if the provided function returns true. Allow functions will typically pull out information on the
```
sfc-gh-jallie, line 38: I'm having trouble here understanding how the 'allow' will work in practice. I don't see a call site other than in tests, where the allow function is trivial.

It feels kinda like an authz check outside the OPA policy, but maybe I'm thinking about it wrong.

sfc-gh-jallie (commented):
looks mostly good. One question re: excessive context creation

proxy/server/server.go
stvnrhodes: Good call, added a boolean. We should have good guarantees that the authz hook has run before any message is received by this channel.

auth/opa/proxiedidentity/proxiedidentity.go
stvnrhodes: I'd put it here to clarify what the bool is for.

After looking at this again, I think it's simpler if we return only the pointer and use "pointer is nil" instead of !ok. Changed and removed the return value names.

auth/opa/proxiedidentity/proxiedidentity.go
stvnrhodes: There's a more concrete example in sansshell-server main in https://github.com/Snowflake-Labs/sansshell/pull/374

```
allowProxyToImpersonate := func(ctx context.Context) bool {
	peer := rpcauth.PeerInputFromContext(ctx)
	if peer == nil {
		return false
	}
	return peer.Cert.Subject.CommonName == *proxyIdentity
}
```

I was considering making it part of the authz policy. It'd look something like

```
package sansshell.authz
default authz = false
authz {
    allow
    not deny
}
deny {
  input.metadata["proxied-sansshell-identity"]
  not input.peer.principal.id = "proxy"
}
```

My worry was that it might be too easy to add the interceptors without updating the rego policy. Now that I think about it more, the benefit of consolidating authz might outweigh that worry.

stvnrhodes: Changed to recommend using the OPA policy, ptal

sfc-gh-jallie (approved):
LGTM

PR #361, opened 2023-10-20
Add an optional TargetConn to RPCAuthInput

stvnrhodes:
We're finding cases where we want the sansshell proxy to make gRPC calls to the target sansshell server as part of gathering input for policy evaluation. For example, for MPA we want the proxy to call out to the server to check if there's any MPA approval for a request.

The ways of doing this in RPCAuthzHook without adding it to RPCAuthInput get very hacky. It's possible to pull out the hosts's address and dial it, but then you're establishing a separate redundant connection and you need to replicate dial options around things like authentication and metrics.

We can't change the hook interface, so our two options are adding a new field to RPCAuthInput or putting the connection into the context. Putting it into the context is possible but we should be putting "request-scoped data that transits processes and API boundaries" in there and a grpc connection doesn't fit that model. We're left with the new RPCAuthInput field as the best option. This PR introduces the first field in RPCAuthInput that's meant purely for use by RPCAuthzHooks and not for policy evaluation.

Part of https://github.com/Snowflake-Labs/sansshell/issues/346

PR #364, opened 2023-10-26
Add server and client implementations for MPA.

stvnrhodes:
These changes are sufficient for MPA when using a direct connection to the server. Here's a few sample commands you can run in parallel to try it out.

```
go run ./cmd/sansshell-server
go run ./cmd/sanssh -client-cert ./auth/mtls/testdata/client.pem -client-key ./auth/mtls/testdata/client.key -mpa -targets localhost healthcheck validate
go run ./cmd/sanssh -client-cert ./services/mpa/testdata/approver.pem -client-key ./services/mpa/testdata/approver.key -targets localhost mpa approve a59c2fef-748944da-336c9d35
```

I've added some new testdata certs because I'm forbidding cases where approver == requester. I've updated the sansshell server code to allow any request if it's requested by our "normal" client cert and approved by our "approver" client cert.

The output of `-mpa` prints a nonconfigurable help message to stderr while waiting on approval. If the command is already approved, the message won't show up.

```
$ sanssh -mpa -targets localhost healthcheck validate
Multi party auth requested, ask an approver to run:
  sanssh --targets localhost:50042 mpa approve a59c2fef-748944da-336c9d35
Target localhost:50042 (0) healthy`
```

This implements the client and server portion, but not the proxy portion. The proxy part mostly builds on top of what I have here and will take advantage of some other features I'm implementing.

- https://github.com/Snowflake-Labs/sansshell/pull/361 for implementing the proxy equivalent of `ServerMPAAuthzHook()`
- https://github.com/Snowflake-Labs/sansshell/pull/358 for implementing the proxy equivalents of `mpahooks.UnaryClientIntercepter()` and `mpahooks.StreamClientIntercepter()`
- https://github.com/Snowflake-Labs/sansshell/pull/359 so that MPA can use the identity of the caller to the proxy instead of the identity of the proxy.


services/mpa/server/server.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// Time to clear out excessive approvals!
+	for len(s.actions) >= maxMPAApprovals {
+		var oldestID string
+		oldestTime := time.Now()
+		for id, act := range s.actions {
+			if act.lastModified.Before((oldestTime)) {
```
sfc-gh-jallie, line 162: nit: what's with the extra parens around 'oldestTime'?

cmd/sanssh/main.go
```diff
@@ -84,6 +85,7 @@ If port is blank the default of %d will be used`, proxyEnv, defaultProxyPort))
 	verbosity        = flag.Int("v", -1, "Verbosity level. > 0 indicates more extensive logging")
 	prefixHeader     = flag.Bool("h", false, "If true prefix each line of output with '<index>-<target>: '")
 	batchSize        = flag.Int("batch-size", 0, "If non-zero will perform the proxy->target work in batches of this size (with any remainder done at the end).")
+	mpa              = flag.Bool("mpa", false, "Request multi-party approval for changes")
```
sfc-gh-jallie, line 88: bikeshed: other help messages typically refer to "commands" ('changes' implies something about mutation to me)

services/mpa/server/server.go
```diff
@@ -0,0 +1,291 @@
...
+}
+func (s *server) WaitForApproval(ctx context.Context, in *mpa.WaitForApprovalRequest) (*mpa.WaitForApprovalResponse, error) {
+	s.mu.Lock()
+	act, ok := s.actions[in.Id]
+	if !ok {
+		return nil, status.Error(codes.NotFound, "MPA request not found")
+	}
+	s.mu.Unlock()
+	select {
+	case <-act.approved:
```
sfc-gh-jallie, line 235: one minor enhancement you might make here is to periodically recheck that the action is still valid, in case the client has set a very long deadline

e.g
```go
for {
        s.mu.Lock()
	act, ok := s.actions[in.Id]
	if !ok {
		return nil, status.Error(codes.NotFound, "MPA request not found")
	}
	s.mu.Unlock()
        timer := time.NewTimer(5 * time.Minute)
        select {
          case <-act.approved:
            return &mpa.WaitForApprovalResponse{}, nil
          case <-ctx.Done()
             return nil, ctx.Err()
          case <-timer.C:
             // go around the loop again
        }
}
```

sfc-gh-jallie (approved):
LGTM

minor/optional nits

services/mpa/server/server.go
stvnrhodes: Huh, that's a typo that happens to compile. Fixed.

cmd/sanssh/main.go
stvnrhodes: I've changed to "command" and added a bit more explanation of what the flag does

services/mpa/server/server.go
stvnrhodes: Good idea, done. I'm doing once-a-minute checks because checking should be cheap (single lock/unlock)

PR #372, opened 2023-11-15
Remove the special-case for the final batch when batching calls

stvnrhodes:
There's some repetitive logic between these calls and I'm going to add slightly more as part of implementing MPA. We can remove the logic duplication by instead adding a conditional to shorten the size of the final batch.


sfc-gh-elinardi (approved):
Nice! This is much easier to follow.

PR #374, opened 2023-11-15
Implement proxied MPA support

stvnrhodes:
This adds support for MPA requests that go through the proxy, building upon the support that we have for direct MPA requests. I've updated documentation with examples and I've added an integration test.

Some notes on implementation quirks:

- Our telemetry handler passes through the metadata specifying the MPA request id from the proxy to the server, triggering its MPA authz hook if the method matches. This works fine because the server will look at proxied identity information for deciding if the request id matches the stored data.
- Our telemetry handler also passes through the justification metadata. If someone forgets to add the handler that does this, justification isn't stored in the MPA request and the proxy MPA authz hook will fail due to the mismatch between the stored action and the requested action.
- sanssh will wait for all targets to be approved before running the final command on any target.

Fixes https://github.com/Snowflake-Labs/sansshell/issues/346


sfc-gh-jallie (approved):
LGTM

PR #383, opened 2023-11-29
Add utils for packages under SansShell services

sfc-gh-cachan:


services/packages/client/utils.go
```diff
@@ -0,0 +1,45 @@
...
+	var errorList []error
+	for resp := range respChan {
+		if resp.Error != nil {
+			err := fmt.Errorf("can't install package %s on target %s: %w", PackageName, resp.Target, resp.Error)
+			errorList = append(errorList, err)
+		}
+	}
+
+	if len(errorList) == 0 {
+		return nil
```
sfc-gh-ldormoy, line 36: you should return `respChan` instead of nil. We want to get the output of each remote package installation.

The client (not implemented in this codebase) should decide what to do with this output (display or suppress it)

sfc-gh-ldormoy: Nevermind, I changed my mind, the output is not needed.

PR #388, opened 2023-12-11
Refactor fdbmovedata with more concurrency

stvnrhodes:
This code was originally written to only work with a single caller. It took a lock for the entire call and no other callers could make progress until the lock was released.

This was done for simplicity of implementation, but turned out to be too simple. If the caller of WaitData died, the context would be cancelled and two possible scenarios could happen.

- If the command was writing to only stdout, the stdout goroutine would exit due to the failure to send the return stream. The stderr goroutine would continue to run waiting for some stderr to get written. For long-running commands, the pipe buffer (~64kb) would eventually fill up and the child subroutine would halt and be unable to progress.
- If the command was writing to both stdout and stderr, the client code would exit both goroutines when it tries to send on the stream and it would get to the `cmd.Wait` part. Long-running commands hit the same failure mode.

The server was taking a lock for the entire wait call, so everything got stuck from this.

I've refactored to support a few different features:

- Locks are now taken in a much more fine-grained way. There shouldn't be any locks that last longer than it takes to spawn a process.
- Multiple FDBMoveDataWait are supported. Each call will stream back stdout/stderr. We'll still forget about a command as soon as it finishes.
- Stdout and stderr are collected regardless of whether FDBMoveDataWait is running. We retain a maximum of 1MB of each to prevent this from consuming unlimited memory.

We still don't support running multiple FDBMoveDataCopy commands at once. If we want to do so, all we need to do is remove the check on `len(s.operations)` and update the tests.


sfc-gh-elinardi (approved):
neat!

PR #400, opened 2024-02-09
Add tcpoverrpc service

sfc-gh-elinardi:
This PR introduces a new service `TCPOverRPC`, providing a way to check TCP connectivity from sansshell target(s) to the given host and port without the capability to send arbitrary packets. This limitation is by design to focus on connectivity verification while ensuring it cannot be misused for sending data or exploited for unintended network traffic generation.  

Testing
```
$ sanssh --targets=localhost tcpoverrpc validate --hostname example.come --timeout 5 4434
TCP validate from localhost (index 0) to example.come: returned error: rpc error: code = Unknown desc = failed to dial tcp example.come:4434: dial tcp: lookup example.come: no such host

$ sanssh --targets=localhost tcpoverrpc validate --hostname example.com --timeout 5 4434 
TCP validate from localhost (index 0) to example.com: returned error: rpc error: code = Unknown desc = failed to dial tcp example.com:4434: dial tcp 93.184.216.34:4434: i/o timeout

$ sanssh --targets=localhost tcpoverrpc validate --hostname example.com --timeout 5 443
TCP validate from localhost (index 0) to example.com: OK
```


services/tcpoverrpc/tcpoverrpc.proto
```diff
@@ -0,0 +1,39 @@
...
+  rpc Ok(HostTCPRequest) returns (google.protobuf.Empty) {}
+}
+
+message HostTCPRequest {
+  // Hostname can be specified as either an ip address or domain name.
+  string hostname = 1;
+  // The destination port.
+  int32 port = 2;
+  // Maximum amount of time (in seconds) to wait for a TCP connect to complete.
+  int32 timeoutSeconds = 3;
```
sfc-gh-bxin, line 38: style: timeout_seconds

services/tcpoverrpc/server/server.go
```diff
@@ -0,0 +1,67 @@
...
+
+	timeout := time.Duration(req.TimeoutSeconds) * time.Second
+	if req.TimeoutSeconds <= 0 {
+		timeout = defaultTimeoutSeconds * time.Second
+	}
+
+	dialer := net.Dialer{Timeout: timeout}
+	conn, err := dialer.DialContext(ctx, "tcp", host)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial tcp %s: %v", host, err)
```
sfc-gh-bxin, line 53: For grpc, it's good to return the standardized error codes:  https://grpc.github.io/grpc/core/md_doc_statuscodes.html;  so here: DEADLINE_EXCEEDED.

See: "google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"

services/tcpoverrpc/client/client.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/Snowflake-Labs/sansshell/client"
+	pb "github.com/Snowflake-Labs/sansshell/services/tcpoverrpc"
+	"github.com/Snowflake-Labs/sansshell/services/util"
+)
+
+const subPackage = "tcpoverrpc"
+
+func init() {
+	subcommands.Register(&tcpCmd{}, subPackage)
```
sfc-gh-bxin, line 38: Just some thoughts on how to structure this functionality in the `sanssh` cmd/subcmd hierarchy....

I know we have a httpoverrpc service, but I feel your tcpoverrpc differs from that model;  httpoverrpc is still interacting with the target host itself.  While your tcpoverrpc is trying to interact with other hosts from sanshell target host.  Based on that, I wonder if we can structure this differently, for example:

1.   as a subcommand of the existing "healthcheck" cmd?
2.  or, start a new top cmd "outbound" or "networking", subcmd: "tcping", to make it more clear the action;  also "outbound" implies more ways to extend with subcommands, say "udpping", or many other protocol layer health checks, icmp/ping, arp etc...

services/tcpoverrpc/client/client.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (p *validateCmd) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&p.Hostname, "hostname", "localhost", "ip address or domain name to specify host")
+	f.IntVar(&p.TimeoutSeconds, "timeout", 10, "maximum amount of time (in seconds) to wait for a TCP connect to complete.")
+}
+
+func (p *validateCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interface{}) subcommands.ExitStatus {
+	state := args[0].(*util.ExecuteState)
+	c := pb.NewTCPOverRPCClientProxy(state.Conn)
+	port, err := strconv.Atoi(f.Arg(0))
```
sfc-gh-bxin, line 90: Nit:  a little weird that 'hostname' is a flag, and port is not...

services/tcpoverrpc/server/server.go
sfc-gh-bxin: Another option is to actually define a response proto type, to return the tcp ping results, normally.  (Think about it:  from SansShell perspective, the code runs successfully in this case; so all error codes are inapproporiate;  maybe we can use a error code indicate client error (like a http status code 408?, which doesn't have a good grpc error mapping;  maybe INVALID_ARGUMENTS? but even that is not optimal)

PR #418, opened 2024-04-05
Deprecate ServerProxiedIdentity*Interceptor

stvnrhodes:
I added these in https://github.com/Snowflake-Labs/sansshell/pull/359 in the original proxiedidentity implementation. In my initial draft I was using them as an additional authorization layer, but I switched to relying on the existing OPA policy authorization layer before submitting.

Now, several months later, I've found another place where I want to use proxied identity data. The interceptors force folks to be intentional about allowing proxying, but in retrospect I think that adds friction without meaningfully helping security. I'd rather remove the need for them.

I'm turning these interceptors into no-ops so that any existing references will continue to work. I've changed proxiedidentity.FromContext to directly work on grpc metadata.


sfc-gh-jallie (approved):
LGTM

PR #443, opened 2024-06-03
Add support for service action reload

sfc-gh-daspencer:
Adds support for reloading a systemd/unknown service via an `ACTION_RELOAD` service action request.


services/service/client/utils.go
```diff
@@ -155,3 +155,21 @@ func DisableRemoteService(ctx context.Context, conn *proxy.Conn, system pb.Syste
...
 }
+
+// ReloadRemoteService is a helper function for reloading a service on a remote target
+// using a proxy.Conn. If the conn is defined for >1 targets this will return an error.
+func ReloadRemoteService(ctx context.Context, conn *proxy.Conn, system pb.SystemType, service string) error {
+	if len(conn.Targets) != 1 {
+		return errors.New("ReloadRemoteService only supports single targets")
+	}
+
+	c := pb.NewServiceClient(conn)
```
sfc-gh-elinardi, line 166: I know all existing functions on this file don't support >1 targets, but it'd be convenient if new functions start supporting it.
Could you please make this call `NewServiceClientProxy` + `ActionOneMany`, and remove the restriction on L162?
Example: https://github.com/Snowflake-Labs/sansshell/blob/92e023d88184d9d4c05be077c695ae0e13263d7c/services/healthcheck/client/utils.go#L27

sfc-gh-elinardi (approved):
LGTM, just one minor comment

services/service/client/utils.go
sfc-gh-daspencer: I've updated the PR with this

sfc-gh-elinardi: Could you also remove the comment on https://github.com/Snowflake-Labs/sansshell/pull/443/files#diff-f4db6bbc237d39cffe696cbadba013e36ceca9e6c7686fcf1726f96228bf154aR160 please? :)

sfc-gh-daspencer: Oh yes

PR #449, opened 2024-06-26
Downgrade Prometheus to v1.0.0 to pull in indirect dependency go-grpc-middleware/v2 at v2.0.0-rc.3

sfc-gh-mtatoczenko:
In an effort to unblock some internal dependency issues with upgrading Sansshell, I am downgrading prometheus to a version that uses go-grpc-middleware/v2 at v2.0.0-rc.3. This should avoid a dependency upgrade that breaks for our internal use case.


sfc-gh-srhodes (approved):
LGTM, the upgrade was a dependabot upgrade and there aren't new features or bug fixes we require in the newer version.

PR #460, opened 2024-07-22
Tweak debug_redact behavior

sfc-gh-srhodes:
In sansshell we use redact to avoid logging sensitive info primarily for avoiding leaking secrets, not to provide privacy. I've switched our redaction from `--REDACTED--` to `REDACTED-<sha256hash>` so that it's possible to figure out if a field matches a known value while not being possible to learn the value purely from reading the field. The empty value is left unredacted because it should never be secret. I've also added support for bytes in addition to strings.

Now that we have a more generous redaction, I've added it to a couple more fields that seem like they might accidentally contain secrets.


auth/opa/rpcauth/redact.go
```diff
@@ -106,15 +107,32 @@ func redactNestedField(message protoreflect.Message, descriptor protoreflect.Fie
...
 			for i := 0; i < list.Len(); i++ {
-				list.Set(i, protoreflect.ValueOfString("--REDACTED--"))
+				if val := list.Get(i).Bytes(); val != nil {
+					list.Set(i, protoreflect.ValueOfBytes([]byte(fmt.Sprintf("REDACTED-%x", sha256.Sum256(val)))))
+				}
 			}
 		}
 	} else {
-		// other than string, clear it
+		// other than string or bytes, clear it
```
sfc-gh-jallie, line 135: fine to leave for later, but I wonder if we'll also want to eventually cover things like StringValue and BytesValue

proxy/testdata/testservice.proto
```diff
@@ -23,6 +23,7 @@ option go_package = "github.com/Snowflake-Labs/sansshell/proxy/testdata";
 message MyNested {
```
sfc-gh-jallie, line 23: might be also nice to have a 'oneof' where one of the options is sensitive. You're looping over the message fields with 'Range', so the implementation is correct, but it would be nice to have a test that exercises this behavior

sfc-gh-jallie (approved):
LGTM

auth/opa/rpcauth/redact.go
sfc-gh-srhodes: Possibly, I've added a TODO. It might be good to treat a message-type field marked with `debug_redact` as though all of the message's fields had `debug_redact`.

The failure mode of not adding more cases here is "information is deleted" instead of "information is printed", so that seems like a safe failure mode if we try annotating a field not covered by our explicit cases.

proxy/testdata/testservice.proto
sfc-gh-srhodes: Added, and I've added a couple of other fields to test code paths that weren't yet being exercised by the tests.

PR #474, opened 2024-08-16
Change packages list command output to satisfy NEVRA format

sfc-gh-ikryvanos:
I had changed `packages list --nevra` logic to use the NEVRA format. Also I extract it to separate domain function to be able to reuse across service and cover with tests

## Is it was tests
I proves that old `search` implementation and new `search`/`list` print package versions in the same format
I proves that on remote machine it is possible to get name-version from list and use it in update

### DIFF
--- a/services/packages/client/client.go
+++ b/services/packages/client/client.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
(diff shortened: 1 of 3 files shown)


sfc-gh-srhodes (approved):
Thanks for the improvement!

sfc-gh-elinardi (approved):

PR #475, opened 2024-08-19
Add pre-commit hooks to fix styling

sfc-gh-ikryvanos:
Add several pre-commit hooks to avoid file formatting comments on code review


sfc-gh-srhodes (approved):
Thanks!

PR #489, opened 2024-09-24
SNOW-1665681 Add set/get operators

sfc-gh-ikryvanos:
Add operations to get and set values in `dotenv` and `yml` files


sfc-gh-elinardi:
Could you please justify the need to split this into controller and usecase? IMO that'd complicate things a bit

services/localfile/localfile.proto
```diff
@@ -275,3 +281,38 @@ message SymlinkRequest {
 message MkdirRequest {
   FileAttributes dir_attrs = 1;
 }
+
+// FileFormat enum what represents file formats
+enum FileFormat {
+  UNKNOWN = 0;
```
sfc-gh-elinardi, line 287: what's the use case of this? isn't the file format must be known / defined by client when making the request?

services/localfile/localfile.proto
```diff
@@ -68,6 +68,12 @@ service LocalFile {
 
   // Mkdir create a new directory.
   rpc Mkdir(MkdirRequest) returns (google.protobuf.Empty) {}
+
+  // Get property from a file of specified type by provided key
```
sfc-gh-elinardi, line 72: nit: property doesn't seem to be the right word. IMO property sounds like file metadata

services/util/array-utils/array.utils.go
```diff
@@ -0,0 +1,37 @@
...
+   "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+   KIND, either express or implied.  See the License for the
+   specific language governing permissions and limitations
+   under the License.
+*/
+
+package array_utils
+
+// IsInArray checks if value is in array
+func IsInArray[T comparable](arr []T, value T) bool {
```
sfc-gh-elinardi, line 20: I can't seem to find any code that's using this function. Is it needed?

services/util/array-utils/array.utils.go
```diff
@@ -0,0 +1,37 @@
...
+	for _, v := range arr {
+		if v == value {
+			return true
+		}
+	}
+	return false
+}
+
+// FindIndexBy finds the index of value in array
+func FindIndexBy[T comparable](arr []T, predicate func(val T) bool) int {
```
sfc-gh-elinardi, line 30: I'd try to avoid having `utils` packages. This function is only used in one place. Should we just place this in the same file, to simplify?

services/localfile/client/infrastructure/input/file.utils.go
```diff
@@ -0,0 +1,39 @@
...
+	"path/filepath"
+)
+
+func getFileTypeFromPath(filePath string) (pb.FileFormat, error) {
+	fileExt := filepath.Ext(filePath)
+
+	switch fileExt {
+	case ".yml":
+		return pb.FileFormat_YML, nil
+	case ".yaml":
```
sfc-gh-elinardi, line 32: you can do `case ".yml", ".yaml"` instead

sfc-gh-elinardi:
Please fix the copyright year

services/localfile/localfile.proto
sfc-gh-ikryvanos: This is required to backward capabilities. If in new version we will add new type, for sansshell server with old schema version new type would be parsed as UNKNOWN and user get correct error

services/util/array-utils/array.utils.go
sfc-gh-ikryvanos: How I see this 2 options. I'm try to avoid compressing and relay only on absolute facts

If we keep util
*Pros*
- code in a separate module are easy to read
- it logically placed in `array-utils` to simplify finding it
- function covered by test and we sure it have no bugs
- it generic and ready to be reused in other cases without modifications

*Cons*
- extra module

If we not extract it:
*Pros*
- no extra module

*Cons*
- next time when similar usecase appears, it would be required to remember that there was first usecase and refactor it to use generic function
- source module code would complicated to understand and module would look massy, new contributor would need first understand what function is belongs to functionality and what to utils and only after that make required changes

------
Besides I'm not see a big difference would the function be in separate module or in the same one.
I now that a little bit tricky, and could lead to extracting every single pice of code to separate module. But if we keep utils application/domain agnostic everything should be ok. Simple test is that it should be easy to copy function as it is any other app and start use it in 2-3 different usecases. For this one is a ton of search usecases were required custom/complicated logic

services/localfile/client/infrastructure/input/file.utils.go
sfc-gh-ikryvanos: done

services/util/array-utils/array.utils.go
sfc-gh-ikryvanos: No it not in used. It was used in one of intermediate solutions. Removing it for now

services/localfile/localfile.proto
sfc-gh-ikryvanos: Updated

sfc-gh-ikryvanos:
Rename branch. New PR link https://github.com/Snowflake-Labs/sansshell/pull/492

PR #491, opened 2024-10-03
SNOW-1665681 Add host overwrite config

sfc-gh-ikryvanos:
Add possibility to overwrite host on making requests


sfc-gh-elinardi (changes requested):
Could you please add more explanations about the goal of this change? i.e. What does this enable the httpoverrpc action to do?

What's the difference between using `WithHostsOverwrite` versus setting the hostname in the url (which gets processed here https://github.com/Snowflake-Labs/sansshell/pull/491/files#diff-0b1b121a0ac0c05fb4cf1d3966b83871d09dc56d844835cf59040de837ddd136R182)?

PR #492, opened 2024-10-04
Add command to get and set data in files of specific formats

sfc-gh-ikryvanos:
Add operations to get and set values in dotenv and yml files


services/util/array-utils/array.utils.go
```diff
@@ -0,0 +1,27 @@
...
+   "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+   KIND, either express or implied.  See the License for the
+   specific language governing permissions and limitations
+   under the License.
+*/
+
+package array_utils
+
+// FindIndexBy finds the index of value in array
+func FindIndexBy[T comparable](arr []T, predicate func(val T) bool) int {
```
sfc-gh-elinardi, line 20: continuing on https://github.com/Snowflake-Labs/sansshell/pull/489#discussion_r1781441781

my preference is to avoid creating new Go packages until there's an actual use case for it.

Having more packages would make the codebase harder to follow.

Also, https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

services/localfile/client/cli-controllers/file.utils.go
```diff
@@ -0,0 +1,37 @@
...
+
+	Unless required by applicable law or agreed to in writing,
+	software distributed under the License is distributed on an
+	"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+	KIND, either express or implied.  See the License for the
+	specific language governing permissions and limitations
+	under the License.
+*/
+
+package cli_controllers
```
sfc-gh-elinardi, line 18: I'd recommend keeping these in client.go to keep it consistent with the other localfile commands.

services/localfile/localfile.proto
```diff
@@ -275,3 +281,38 @@ message SymlinkRequest {
 message MkdirRequest {
   FileAttributes dir_attrs = 1;
 }
+
+// FileFormat enum what represents file formats
+enum FileFormat {
+  UNKNOWN = 0;
```
sfc-gh-elinardi, line 287: https://github.com/Snowflake-Labs/sansshell/pull/489#discussion_r1781431167
continuing on this.

You mentioned that the UNKNOWN file format is for the server to tell clients that it doesn't recognize the file format.
However, I don't see any server code referencing this unknown file format. How would this be used by the server?

sfc-gh-ikryvanos: It for a case when we use new client version with new FileFormat, e.g. json, but old server. In this particular case it value would be casted to UNKNOWN and server reply that it doesn't support this format

services/localfile/client/cli-controllers/file.utils.go
sfc-gh-ikryvanos: It cause too much code in client, for me it hard to follow and hard to find definition of exact command what I need. By keep it in separate file I:
- improve readability
- simplify search of the particular cli handler definition

Have 1 component per file it common best practice for a lot of languages. 

What the advantage of keeping code in client.go approach?

services/util/array-utils/array.utils.go
sfc-gh-ikryvanos: > Having more packages would make the codebase harder to follow.

But have the code in place make it harder to understand. It is much simpler to undestand what is going on when you see
```golang
index := arrUtils.FindIndexBy(arr, func (val) => val == 5)
// vs
index := -1
for i, val := range arr {
  if val == 5 {
     index := i
  }
}
```
`FindIndexBy` is super generic and I definitely sure that I reuse it

sfc-gh-ikryvanos:
`Usecase` it is a component with single entry point what handle pure business logic of dedicated business process. In this PR it is logic to get/set value from/to a particular file by specific key. Mostly used in clean architecture, but event without following all "clean" principal is very handy. Main advantage to use it here is see whole process logic in single file and ability to cover it with unit tests

`factory` its a pattern what allows to get specific implementation of particular interface, and keep code independent from this specific implementation. `repository` its an adapter to specific storage which incapsulate all specific of this store inside and provide simple interface outside. Together is allows me to incapsulate all logic specific to particular file type inside of repo and use only common interface. With this approach we minimize code changes required to add new type of file. 

Also separation data access logic from business logic allows to cover it with integration tests with all very specific tests cases for a particular implementation. And cover business logic with unit tests. It allows to reduce number of tests what would reduce effort on tests support. If we not do this it would require to cover both implementation with very simular e2e tests with almost the same preparation logic, e.g.:
- should reply with error in case file not exists
- should reply with error if file impossible to lock
- should reply with error if file fails with error

> Why do we need them to be in a separate package?

It allows to make interface implementation private and avoid it direct usage outside of a module, besides it keep module namespace cleaner.

 > And why are they located under infrastructure/output? What does the file path indicate?

It is take from hexagonal architecture. It core concept is separate app into 3 main parts 
- input infrastructure, adapters what receive input signals from app users and handle all client interface specific logic. Component examples: api endpoint handlers, cli controller
- output infrastructure, very simpler to input but from the different side, adapters from business logic to some 3rd party api, db, fs and etc. Component examples: repository, gateway, client
- application. Logic of application itself independent from interface specific and used external resources

hexagonal architecture was taken as good reference simple enough to be easily used by any developer, but powerful enough  to give good standard which allow to scale app functionality without make it massy

services/localfile/client/cli-controllers/file.utils.go
sfc-gh-elinardi: I'd say keeping it in client.go is much more simpler approach, and wouldn't introduce new variances. I don't see anything wrong with long, but well-structured, file. On the other hand, breaking it into many files would make it harder to navigate.
Also if we decide to refactor, let's do it for all other commands.

services/localfile/localfile.proto
sfc-gh-elinardi: could you explain more about how the server would reply with UNKNOWN? I don't see any references to `FileFormat_UNKNOWN` in the server code.

sfc-gh-elinardi:
I guess in general what I'm trying to understand is: what are the problems with the current codebase that require major changes to its structure? 
I'd try to avoid introducing major changes to the code structure unless it's really necessary. 
Perhaps you could help me understand these problems by sharing some code snippets.

services/util/array-utils/array.utils.go
sfc-gh-ikryvanos: I see several problems with this approach, especially when we scale it to each situation:
- when it would require to reuse it in different module it would require to move function and tests what would increase code changes in PR
- for me it would require from me to remember all places where I make some handy utils to being able to reuse them
- for other devs, highly likely, they won't even know that such utility existing and implement their own. In this case there 2 sub scenario:
   - I need to reject on code review and force them to refactor some code, what increase task scope and delivery time
   - we would have a lot similar function across project, but different enough to make refactoring. Together with other factors it increase chance of regression

On the other hand I don't see solid benefits to not doing this

services/localfile/client/cli-controllers/file.utils.go
sfc-gh-ikryvanos: It subjective. For me it harder to navigate in big file. When I see clear logical structure of the whole project in much easier for me. Mostly it matter of habit, I suggest to try reference only solid facts and measurable pros and cons for a project.

solid fact from my side:
- devs would work with single command at a time. Other commands code not necessary for them
- to find exact command devs spend less time if it in separate files, lets have an example, lets assume we need to find command controller, but we not remember its exact name:
   - right now:
      - open relevant service
      - open client.go
      - scroll thought a hundreds lines of code to find command controller definition
    - with 1 command per file
      - open relevant service
      - open folder with commands handlers
      - go thought <20 file names to find command controller definition

services/localfile/localfile.proto
sfc-gh-ikryvanos: I was wrong. I was pretty sure that if I send from a client enum value what not exists on a server (case when client has newer version, with new enum value, then server) GRPC parser would automatically cast it to 0 and determine it as unknown. There is not the case, I'll take this in account next time we would need to add prover validation for enums to avoid such situations

But it still useful to have UNKNOWN value in a enum. e.g. for a cases when we need zero value. e.g. https://github.com/Snowflake-Labs/sansshell/blob/d15e4f94d90fa383198b76aa32252be07c527ce6/services/localfile/client/cli-controllers/file.utils.go#L35. Otherwise I would need to return some enum zero value e.g. FileFormat_YML. I could remove it, but it might be risky

sfc-gh-ikryvanos:
> I'd try to avoid introducing major changes to the code structure unless it's really necessary.

Really necessary is subjective. For me it really necessary, but lets try to avoid subjective arguments and point to solid and measurable points:
- currently the only way to test functionality is integration tests. Integration tests has their own pros and cons, the most important one:
   - covering with integration tests would ends up with too long execution and very costly refactoring, which is possible to avoid
   - impossible to mock dependency to check specific cases. In this particular PR there is no difference why exactly is impossible to get value from a file, behavior would be the same and response would be almost the same. But with integration tests I would need to cover it all to be sure it all works
   - each test need to have similar preparation steps, like set up grpc server and etc
   - *subjective* low level code to access external resources such as file system, http api, fdb mixed with parameter validation logic. I need more time to understand it. Logic separation gives opportunity not keep in a context only things important right now, which based on my experience gives higher effectiveness and less bugs
   - long and subjective code review. It not blaming, we need to pass through alignment phase to have effective collaboration. Right now any pice of code, any used pattern would be challenged by reviewer point of view. There is no any reason for me to create new function/method/utility or not to create, it would be measured just by reviewer opinion from his point of view. It rise discussion about do we need to separate something to different module or not, is it ok to extract something to different function or not. Each discussion takes time. Clear structure and clear rules allows to focus on discussion of only really important things like like proto schema, cli commands arguments, backward capabilities and etc 

What it gives:
- clear test strategy which gives high level quality
- clear rules about how to structure you code
- less code changes
- less discussion
- higher developer effectiveness
- clear way how to scale functionality up to infinity without necessity to refactor whole project

sfc-gh-elinardi (approved):

PR #493, opened 2024-10-04
Add possibility to overwrite host in a client 

sfc-gh-ikryvanos:
Add possibility to overwrite host on making requests


sfc-gh-ikryvanos:
Question from @sfc-gh-elinardi 

> Could you please add more explanations about the goal of this change? i.e. What does this enable the httpoverrpc action to do that's not possible today?
>
> What's the difference between using WithHostsOverwrite versus setting the hostname in the url (which gets processed here https://github.com/Snowflake-Labs/sansshell/pull/491/files#diff-0b1b121a0ac0c05fb4cf1d3966b83871d09dc56d844835cf59040de837ddd136R182)?


It is required to being able to create HTTP transport with specifc config and being able to provide it as transporter to custom http client. It is required 3rd party system

sfc-gh-srhodes (commented):
Could you add a test for this? Maybe launch a http server, make a call that sets the host, and validate that the http server got a Host header with the expected value?

sfc-gh-ikryvanos:
@sfc-gh-srhodes covered with tests

sfc-gh-elinardi (approved):

sfc-gh-srhodes (approved):

PR #494, opened 2024-10-07
Expose Unix Domain Socket from server for local access

sfc-gh-srudnicki:
Apart from providing access to a node from the outside, Sansshell
could conceivably be viewed as an interface between services running
on the node and the node's operating system. For example, instead of
adding a sudoers entry to allow some operation which would normally
require root privileges, we could expose the operation as a Sansshell
gRPC method, and call it from the inside: from some process which runs
on the same node as the Sansshell server. This way, we would gain
fine-grained authorization and auditability.
    
To make this approach possible, we modify the Sansshell server to
optionally expose a Unix domain socket which offers the same gRPC
methods as the main TCP server. The server which handles the socket
does not require TLS authentication, because obtaining the proper
certificates may be difficult for local clients. This should be OK,
considering how MITM attacks using the socket would only be possible
if the node has already been compromised to some extent.
    
Unix permissions/ACLs on the socket file can be used as a security
measure. The server makes this possible by exposing a hook function
which can configure the socket's ownership and permissions once the
socket is created. This hook mechanism may be used in the future to
specify socket permissions and ownership on the server command line.
    
In the rego authentication rules, clients using the Unix socket can be
identified by having `input.peer.net.network = "unix"`.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -20,11 +20,14 @@ package server
 import (
 	"context"
 	"fmt"
+	"io/fs"
 	"net"
+	"os"
 
 	"github.com/go-logr/logr"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/stats"
 
 	"github.com/Snowflake-Labs/sansshell/auth/opa"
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -51,6 +47,7 @@ type serveSetup struct {
 	streamInterceptors []grpc.StreamServerInterceptor
 	statsHandler       stats.Handler
 	services           []func(*grpc.Server)
+	postStartHook      func(*grpc.Server)
```
sfc-gh-ikryvanos, line 50: nit: Usually hook is something sync for what we wait for result, e.g. webhooks. From the usage it is looks like observer pattern, like you just notify some listeners that even happends.

I suggest to
```suggestion
	onPostStartListeners      []func(*grpc.Server)
```

sfc-gh-ikryvanos (approved):

server/server.go
sfc-gh-srudnicki: Sure, that makes sense.

PR #499, opened 2024-10-15
Add flag to ensure folder exists before copy

sfc-gh-ikryvanos:
Add flag to ensure destination folder exists to being able to copy file to unexisted folder in one command


services/localfile/server/localfile_test.go
```diff
@@ -1519,53 +1519,37 @@ func TestCopy(t *testing.T) {
 	t.Cleanup(func() { conn.Close() })
 
 	// Setup source bucket/file and contents
-	contents := "contents"
```
sfc-gh-ikryvanos, line 1522: remove file creation because it makes tests depends on each other

services/localfile/server/localfile_test.go
```diff
@@ -1519,53 +1519,37 @@ func TestCopy(t *testing.T) {
...
-	temp := t.TempDir()
 	uid, gid := os.Getuid(), os.Getgid()
-	f1, err := os.CreateTemp(temp, "file")
-	testutil.FatalOnErr("CreateTemp", err, t)
-	if n, err := f1.WriteString(contents); n != len(contents) || err != nil {
-		t.Fatalf("WriteString failed. Wrote %d, expected %d err - %v", n, len(contents), err)
-	}
-	err = f1.Close()
-	testutil.FatalOnErr("Close", err, t)
```
sfc-gh-ikryvanos, line 1531: remove file creation because it makes tests depends on each other

services/localfile/server/localfile_test.go
```diff
@@ -1576,134 +1560,236 @@ func TestCopy(t *testing.T) {
...
 				},
 				Bucket: "file://not-a-path",
-				Key:    filepath.Base(f1.Name()),
+				Key:    "some-source-file",
 			},
-			wantErr: true,
 		},
-		{
-			// Same here. We don't need to test all finalization permutations as TestWrite did that.
-			name: "Fail due to overwrite = false",
```
sfc-gh-ikryvanos, line 1612: extract it to separate test, because it was too hard to keep so different cases in the same test

services/localfile/server/localfile_test.go
```diff
@@ -1576,134 +1560,236 @@ func TestCopy(t *testing.T) {
...
-			filename:  filepath.Join(temp, "/file"),
-			wantErr:   true,
-			touchFile: true,
-		},
+			Bucket: fmt.Sprintf("file://%s", filepath.Dir(sourceFile.Name())),
+			Key:    filepath.Base(sourceFile.Name()),
+		}
 
-		{
-			name: "Copy a file",
```
sfc-gh-ikryvanos, line 1645: extract it to separate test, because it was too hard to keep so different cases in the same test

services/localfile/server/localfile_test.go
```diff
@@ -1576,134 +1560,236 @@ func TestCopy(t *testing.T) {
...
+			// ACT
+			_, err := client.Copy(context.Background(), tc.req)
+
+			// ASSERT
+			t.Log(err)
+			testutil.WantErr(tc.name, err, true, t)
+		})
+	}
+
+	t.Run("Copy a file", func(t *testing.T) {
```
sfc-gh-ikryvanos, line 1606: moved copy file test

services/localfile/server/localfile_test.go
```diff
@@ -1576,134 +1560,236 @@ func TestCopy(t *testing.T) {
...
+		destinationContent, err := os.ReadFile(destinationFilename)
+		testutil.FatalOnErr("ReadFile()", err, t)
+
+		if got, want := destinationContent, []byte(expectedDstFileContent); !bytes.Equal(got, want) {
+			t.Fatalf("contents not equal.\nGot : %s\nWant: %s", got, want)
+			return
+		}
+	})
+
+	t.Run("Copy a file to not existed directory with ensure destination folder path", func(t *testing.T) {
```
sfc-gh-ikryvanos, line 1670: new test for new func

services/localfile/server/localfile_test.go
```diff
@@ -1576,134 +1560,236 @@ func TestCopy(t *testing.T) {
...
+		testutil.FatalOnErr("ReadFile()", err, t)
+
+		if got, want := destinationContent, []byte(expectedDstFileContent); !bytes.Equal(got, want) {
+			t.Fatalf("contents not equal.\nGot : %s\nWant: %s", got, want)
+			return
+		}
+	})
+
+	// Same here. We don't need to test all finalization permutations as TestWrite did that.
+	t.Run("Fail due to overwrite = false", func(t *testing.T) {
```
sfc-gh-ikryvanos, line 1735: moved "Fail due to overwrite = false" test

services/localfile/client/client.go
```diff
@@ -978,14 +978,15 @@ func (p *lsCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interface{
...
 type cpCmd struct {
-	bucket    string
-	overwrite bool
-	uid       int
-	username  string
-	gid       int
-	group     string
-	mode      int
-	immutable bool
+	bucket          string
```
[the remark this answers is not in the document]
sfc-gh-ikryvanos: Bucket is source file path. For a context it could be `file:///some-abs-path/`, `s3://s3-bucket/key/prefix/` and etc. It not influence on copy behavior

services/localfile/server/localfile.go
```diff
@@ -493,6 +493,15 @@ func (s *server) Copy(ctx context.Context, req *pb.CopyRequest) (_ *emptypb.Empt
 	}
 	filename := a.Filename
 	logger.Info("copy file", filename)
+
+	if d.EnsureOutputDirectory {
+		outputDir := filepath.Dir(filename)
+		if err := os.MkdirAll(outputDir, os.ModePerm); err != nil {
```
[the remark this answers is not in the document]
sfc-gh-ikryvanos: Because we need to be sure that process would be able to write file. File would have its own modeperm and it would be save enough

services/localfile/client/client.go
```diff
@@ -1003,6 +1004,7 @@ sure to use a fully formed directory. i.e. copying /etc/hosts would be --bucket=
 func (p *cpCmd) SetFlags(f *flag.FlagSet) {
 	f.StringVar(&p.bucket, "bucket", "", "If set to a valid prefix will copy from this bucket with the key being the source provided")
 	f.BoolVar(&p.overwrite, "overwrite", false, "If true will overwrite the remote file. Otherwise the file pre-existing is an error.")
+	f.BoolVar(&p.ensureOutputDir, "ensure-output-dir", false, "If true will ensure output dir exists before copy. Otherwise the if folder not exists copy would fail.")
```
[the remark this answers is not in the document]
sfc-gh-ikryvanos: Change flag name and change description

services/localfile/client/client.go
```diff
@@ -978,14 +978,15 @@ func (p *lsCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interface{
...
-	overwrite bool
-	uid       int
-	username  string
-	gid       int
-	group     string
-	mode      int
-	immutable bool
+	bucket               string
+	overwrite            bool
+	createDestinationDir bool
```
sfc-gh-ylefloch, line 983: I can't help but notice that the style of those options is usually one word long, pretty terse. Here a flag with a longer name made of multiple words is added - which to me looks a bit like a red flag.

Additionally, in Unix systems, a different command is used to create directories and it's not an option for `cp`.

I haven't worked on this repo before, so it's hard for me to tell whether this is in line with its usual style. This is where I would like to defer to the maintainers.

services/localfile/server/localfile.go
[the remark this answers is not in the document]
sfc-gh-ylefloch: This doesn't look like the right permissions to me, since `os.ModePerm` is a special member of the enum with all the bits set

services/localfile/client/client.go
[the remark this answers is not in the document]
sfc-gh-ylefloch: I see, thanks

sfc-gh-ylefloch (commented):
One point outstanding on file permission

sfc-gh-elinardi:
Thank you for the effort to improve our user experience! 

I have some thoughts about adding the directory creation functionality to the `copy` action:

1. This is already achievable using the `mkdir` action.
2. Adding directory creation to the `copy` action would give it two responsibilities: copying files and creating directories. I think we should try to keep each action has a single responsibility, to maintain clean separation of concerns.

Instead of doing this, my recommendation is that we should focus on developing a solution that makes running multiple-step actions easier.

Currently, running multiple sansshell actions is only achievable using bash script that invokes `sanssh` repeatedly. It'd be nice if we have a solution that enable users to define something like a "workflow" or "chained operations".

sfc-gh-ikryvanos:
> Currently, running multiple sansshell actions is only achievable using bash script that invokes sanssh repeatedly. It'd be nice if we have a solution that enable users to define something like a "workflow" or "chained operations".

Agreed, this option gives more possibilities for us

PR #511, opened 2024-11-13
Add way to specify user for exec command.

sfc-gh-mwalas:


services/exec/server/exec.go
```diff
@@ -73,6 +87,24 @@ func (s *server) StreamingRun(req *pb.ExecRequest, stream pb.Exec_StreamingRunSe
 	}
 
 	cmd := exec.CommandContext(ctx, req.Command, req.Args...)
+	if req.User != "" {
+		gid, uid, err := resolveUser(req.User)
```
sfc-gh-iyashchyshyn, line 91: These arguments should be swapped actually, its `uid, gid, err`.

sfc-gh-psmolenski (approved):
LGTM

sfc-gh-ikryvanos (commented):
Would you mind to add readme similar to https://github.com/Snowflake-Labs/sansshell/blob/main/services/network/README.md

sfc-gh-ikryvanos (approved):

PR #513, opened 2024-11-15
Add File Shred command

sfc-gh-iyashchyshyn:
Adds `shred` sub-command to `file` command with support for zeroing-pass, file-removal and force-permissions-change flags.


services/localfile/server/shred.go
```diff
@@ -0,0 +1,54 @@
...
+	logger := logr.FromContextOrDiscard(ctx)
+	recorder := metrics.RecorderFromContextOrNoop(ctx)
+
+	if err := util.ValidPath(req.Filename); err != nil {
+		recorder.CounterOrLog(ctx, localfileShredFailureCounter, 1, attribute.String("reason", "invalid_path"))
+		return nil, err
+	}
+
+	// TODO: check if hardlink
+	// TODO: check if not a /dev or other reserved dirs based file
```
sfc-gh-psmolenski, line 46: Is there a requirement saying that `shred` should be doing additional checks for file type and location? `rm` command doesn't seem to implement any such logic.

sfc-gh-iyashchyshyn: I probably have somewhat outdated info, it appears both /dev/mem and /dev/port are both "disabled" by default since 2.6.xx. 

I guess if someone decides to open devices for read/write over ssh, it shouldn't really matter. Maybe someone wants to shred their HDD remotely. So yea I agree we can probably skip this check.

services/localfile/server/shred.go
```diff
@@ -0,0 +1,79 @@
...
+		args = append(args, "-z")
+	}
+
+	if req.Remove {
+		args = append(args, "-u")
+	}
+
+	args = append(args, req.Filename)
+
+	// we want to fail as soon as possible, if the file is unreadable, opened by someone else or something else happens
```
sfc-gh-mwalas, line 66: Was there any alternative you considered here other than attempting shredding and returning error as Internal? I am trying to understand why this comment needs to be in place.

I think what you do here is reasonable, might be little weird that we return "Internal" for filenames that are not found, but we shouldn't go into business of parsing linux commands output(s).

services/localfile/server/shred.go
sfc-gh-mwalas, line 1: Is there a way to reasonably extend the localfile_test.go to cover shredding with some test case? (I guess some test on zero-ing without removal is something we could do)

services/util/util.go
```diff
@@ -456,3 +456,36 @@ func IsStreamToTerminal(stream io.Writer) bool {
 		return false
 	}
 }
+
+// Which attempts to find an executable in PATH and returns its absolute path if successful
+// note that it tries to find the executable anywhere it can, in case of failure to read a directory
+// listed in one of the PATH's components, it will continue its operation and only return an error in case
+// no paths were readable and/or executable couldn't be found in either of them.
+// Note that it does not check whether a file is actually executable by current user.
+func Which(exeName string) (string, error) {
```
sfc-gh-mwalas, line 465: For ansible, the choice so far was to hard-code the bin https://github.com/Snowflake-Labs/sansshell/blob/a76925d925410c2a6bc5e0c2da05669c12d48a4b/services/ansible/server/ansible.go#L39

do you expect us to have different shredding binary specified per user, hence we start to use `which` like approach? If so, we should probably sudo to the user first before resolving path?

(not sure if we need that, and if so we could just start with binding to `/usr/bin/shred`?)

sfc-gh-mwalas (approved):

services/localfile/server/shred.go
sfc-gh-iyashchyshyn: Yup will remove that. The alternative was doing a bunch of checks to fail before actually attempting a shred (and producing better errors), but that's probably overkill in this case.

services/localfile/client/shred.go
```diff
@@ -0,0 +1,81 @@
...
+		for _, e := range state.Err {
+			fmt.Fprintf(e, "All targets - shred client error: %v\n", err)
+		}
+		return subcommands.ExitFailure
+	}
+
+	retCode := subcommands.ExitSuccess
+	for r := range respChan {
+		if r.Error != nil {
+			fmt.Fprintf(state.Err[r.Index], "shred client error: %v\n", r.Error)
```
sfc-gh-ikryvanos, line 76:
```suggestion
			fmt.Fprintf(state.Err[r.Index], "shred failed: %v\n", r.Error)
```

Lets not specify sort of error here, it would be clear out of error message

services/localfile/server/shred.go
```diff
@@ -0,0 +1,79 @@
...
+	// to it while we're trying to shred it, we might as well just fail during command execution
+	r, err := util.RunCommand(ctx, shredPath, args)
```
sfc-gh-ikryvanos, line 68: To run `shred` command we need to be sure it would be on every VM and every base image where sansshell already in use and/or would be in use. 

What do you think can we avoid using it and make go implementation like it done in https://github.com/pabloandresm/shred/blob/main/shred.go instead?

sfc-gh-iyashchyshyn: It would be a little bit more complicated to make a proper implementation for HDDs (for SSDs its of course not going to do anything). I don't know much about magnetic storage, but I think there's something to the whole overwrite pattern, judging by the sources here its using pseudo-random (or optionally "true" random hardware based RNG I guess) seeds and patterns for the passes: https://github.com/coreutils/coreutils/blob/master/src/shred.c#L674

Its usually available on every Linux machine, even most lightweight distros include it - both Wolfi and Alpine.
That being said, maybe we can just fail in this case? Alternatively we can just do an RM (it achieves same thing on SSDs).  

As an alternative I could just add a naive basic implementation as a backup for when actual Shred is not available on the system (as unlikely as that is). 

Frankly I'm not sure if it even makes sense to add Shred though, HDDs are of course still used in Clouds, but chances are whatever is being shredded is going to reside on the boot disk (and its would very uncommon for boot disk to be an HDD imo). Maybe the "solution" here is to discard this whole PR?

sfc-gh-mwalas: I don't believe we can maintain, long term, a full compatibility contract along the lines of what was suggested above ("on every VM and every base image where sansshell already in use and/or would be in use"). It is not realistic to have such a goal, even golang core libraries have quirks of that nature.

Once I say this, I don't feel it is a good time to go full steam on native shred implementation, I would rather stick to use whatever we have in environment until we know we need a native implementation.

services/localfile/README.md
```diff
@@ -130,3 +130,28 @@ Examples:
 # Set file mode of `/opt/hello` to 644
 sanssh --target $TARGET file chmod --mode=644 /opt/hello
 ```
+
+### sanssh file shred
+Shred file using `/usr/bin/shred` (`/opt/homebrew/bin/shred` on Mac OS) utility.
```
sfc-gh-psmolenski, line 135: Missing article - "shred `a` file"

services/localfile/server/shred_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	_, err = client.Shred(ctx, &pb.ShredRequest{
+		Filename: fileName,
+		Force:    false,
+		Zero:     false,
+		Remove:   false,
+	})
+	testutil.FatalOnErr("shred", err, t)
+
+	if getFileContent(t, fileName) == data {
```
sfc-gh-psmolenski, line 53: nit: We should fail the test if the shredded file contains any data, not necessarily the original data.

sfc-gh-psmolenski (approved):

sfc-gh-ikryvanos (approved):

sfc-gh-mwalas (approved):

services/localfile/server/shred_test.go
sfc-gh-iyashchyshyn: Thanks, I changed this to include zero-ing pass and checking that all bytes are zeros after. As a side note Shred will actually extend the file size to `((size_of_original_data / block_size) + 1) * block_size`, typically block size is 512/4096 bytes on HDDs (4k on modern ones), so for a <block_size file it will be extended to 4kb (filled with zeros/random data depending on Shred's settings).

PR #514, opened 2024-11-20
Fix deadlock on return-early on not_found during WaitForApproval 

sfc-gh-mwalas:
In the mpa flow we could end up with servers getting into "close to deadlock" state as we failed to release mutex in one of the branches of the mpa flow.


sfc-gh-psmolenski (approved):

sfc-gh-lgrabowski (approved):
yeah, unlock in mutex in a loop and no way to defer it! Great catch!

PR #517, opened 2024-12-13
Support for multi-arg rm and rmdir

sfc-gh-iyashchyshyn:
Changes `rm` and `rmdir` to accept multiple args (absolute paths) and return a compound error for any dirs/files that failed to be deleted. Also add `-r` flag to `rmdir` which deletes everything inside of directory, along with directory itself. Behaviour in general follows coreutils `rm` with `-r` flag.


services/localfile/localfile.proto
sfc-gh-psmolenski, line 1: It looks like a breaking change. Instead of modifying `filename` field, we could add `filenames` field of type `repeated string` and on server side `filename` value will be used only if `filenames` field is empty.

sfc-gh-iyashchyshyn: I looked into some options and I don't think there's a nice way of doing this for us. Technically, with current code (before the fix below) new-client-old-server scenario would just result in deletion of the last file, with the approach above its going to be first-file deleted only, so I implemented that for now. 

That being said, we have no good way (without connecting that is) of knowing what version of SansShell server we're sending the request to, so we'll have to populate both of these fields at all times basically to ensure that both old and new versions work.

sfc-gh-ikryvanos: Prop remaining also will cause non backward compatible changes of opa policy and could open backdoor to remove what actually should't be removed.

To avoid this lets avoid request shape changes, instead send several remove requests to remove single file/dir

sfc-gh-ikryvanos (changes requested):

PR #518, opened 2024-12-18
Configurable timeout for dmesg scan

sfc-gh-iyashchyshyn:
Limited from 2 to 30 seconds, value is passed in seconds. Also includes some docs for sysinfo.


services/sysinfo/client/client.go
```diff
@@ -131,6 +132,7 @@ func (p *dmesgCmd) SetFlags(f *flag.FlagSet) {
 	f.Int64Var(&p.tail, "tail", -1, "tail the latest n lines")
 	f.BoolVar(&p.ignoreCase, "i", false, "ignore case")
 	f.BoolVar(&p.invertMatch, "v", false, "invert match")
+	f.Int64Var(&p.timeout, "timeout", 2, "timeout collection of messages after N seconds, default is 2 seconds")
```
sfc-gh-iyashchyshyn, line 135: Unsure if it makes any sense to use milli/microseconds here, I specifically opted for int64 in case we want to change it to milliseconds/microseconds in the future. Very unlikely that even milliseconds would make any sense though.

sfc-gh-iyashchyshyn: Thanks, will look into it!

services/sysinfo/server/sysinfo_linux.go
```diff
@@ -195,3 +196,16 @@ var getJournalRecordsAndSend = func(ctx context.Context, req *pb.JournalRequest,
...
+func getTimeout(provided int64) time.Duration {
+	const maxTimeout = 30 * time.Second
+	const minTimeout = 2 * time.Second
+	timeout := time.Duration(provided) * time.Second
+	if timeout > maxTimeout {
+		timeout = maxTimeout
+	}
+	if timeout < minTimeout {
+		timeout = minTimeout
+	}
```
sfc-gh-mwalas, line 209: Since this adjusts silently (without feedback to customers) values on server side I would consider it a defence in depth and have these ranges validated (with error messages) also in sanssh client.

sfc-gh-mwalas (approved):
Consider passing around Duration instead of # of seconds.

services/sysinfo/server/sysinfo_linux.go
```diff
@@ -195,3 +196,16 @@ var getJournalRecordsAndSend = func(ctx context.Context, req *pb.JournalRequest,
 	}
 	return nil
 }
+
+func getTimeout(provided int64) time.Duration {
+	const maxTimeout = 30 * time.Second
```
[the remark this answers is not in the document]
sfc-gh-iyashchyshyn: Multiple reads are not an issues, its the buffering, we can go OOM in theory if the kernel is chatty enough and many clients connect to server and try dmesg at the same time. Its a far fetched scenario given the general size of data here, but still probably worth it.

I improved control logic here a bit too, now at least if the client cancels request/drops connection, the read buffering should be cancelled too. Before my changes it was just going to read kmsg until the timeout was reached. So maybe we can increase this number too now, not quite sure what would make sense though, I think 30s should be plenty for now though.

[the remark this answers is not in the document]
sfc-gh-mwalas: OK, wrt to potential load on sansshell server -- I think we should rewrite this logic, as it is naively now just materializing everything in memory (and does grep later) into streaming. Resolving this for now.

services/sysinfo/client/client.go
```diff
@@ -116,12 +117,13 @@ type dmesgCmd struct {
...
 	ignoreCase  bool
 	invertMatch bool
+	timeout     time.Duration
 }
 
 func (*dmesgCmd) Name() string     { return "dmesg" }
 func (*dmesgCmd) Synopsis() string { return "View the messages in kernel ring buffer" }
 func (*dmesgCmd) Usage() string {
-	return `dmesg [--tail=N] [--grep=PATTERN] [-i] [-v]:
+	return `dmesg [--tail=N] [--grep=PATTERN] [-i] [-v] [--timeout=N]:
```
sfc-gh-psmolenski, line 126: This parameters is now called `dmesg-read-timeout`, not `timeout`

sfc-gh-ikryvanos (approved):

services/sysinfo/server/sysinfo.go
```diff
@@ -115,6 +113,21 @@ func (s *server) Dmesg(req *pb.DmesgRequest, stream pb.SysInfo_DmesgServer) erro
 	return nil
 }
 
+func normalizeTimeout(pbTimeout *durationpb.Duration) time.Duration {
+	if pbTimeout == nil {
+		return pb.MinDmesgTimeout
+	}
+
+	timeout := pbTimeout.AsDuration()
+	if timeout > pb.MaxDmesgTimeout {
```
sfc-gh-psmolenski, line 122: Shouldn't we throw an error if the requested timeout is out of range?

sfc-gh-psmolenski (approved):

services/sysinfo/server/sysinfo.go
sfc-gh-iyashchyshyn: Thanks, replaced normalization with error instead, it will still default to min value for nil param value (older client, newer server) but should throw an error on any value that is out of bounds.

PR #520, opened 2025-01-10
Added option to define PerRPCCredentials

sfc-gh-psmolenski:
Sanssh client can now be configured to use per-RPC credentials. Useful for using OAuth and ID tokens.

### DIFF
--- a/cmd/sanssh/client/client.go
+++ b/cmd/sanssh/client/client.go
@@ -23,6 +23,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"google.golang.org/grpc/credentials"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


sfc-gh-mwalas (approved):
Looks good, but I do prefer to review such changes in context of how they are being used, consider next time merging such additions with usage.

sfc-gh-ikryvanos (approved):

cmd/sanssh/client/client.go
```diff
@@ -77,6 +78,8 @@ type RunState struct {
 	BatchSize int
 	// If true, add an interceptor that performs the multi-party auth flow
 	EnableMPA bool
+
+	credentials.PerRPCCredentials
```
sfc-gh-iyashchyshyn, line 82: Silly question, but how is it passed to RunState from flags/command line?

sfc-gh-iyashchyshyn (approved):

PR #526, opened 2025-01-31
Add API versioning and localfile parallel read with grep

sfc-gh-mwalas:
Allows to read multiple files at once through a wildcard syntax on a directory (limited only to a single trailing `*` for the time being).

This change also introduces a simple API versioning. Reason being:
 - added fields such as `grep` etc do not meaningfully extend the application of the method
 -  however the `/*` syntax allows to skip potential deny rules on reading particular file in subdirectory in rego in a situation where someone allowed reading files in general for a suffix of `/tmp/example` but later added a deny rule for exact path of `/tmp/example/secret`
 
API versioning proposal is covered in `README.md`.


services/localfile/server/localfile.go
```diff
@@ -132,89 +153,153 @@ func (s *server) Read(req *pb.ReadActionRequest, stream pb.LocalFile_ReadServer)
...
-		if err == io.EOF {
-			// If we're not tailing then we're done.
-			if r != nil {
-				break
-			}
-			if err := dataReady(td, stream); err != nil {
-				recorder.CounterOrLog(ctx, localfileReadFailureCounter, 1, attribute.String("reason", "dataready_err"))
+	errs, ctx := errgroup.WithContext(ctx)
+	fileChan := make(chan string, 100)
+	if path, found := strings.CutSuffix(file, "/*"); found {
```
sfc-gh-psmolenski, line 158: Any reason you didn't use `filepath.Glob`?

services/localfile/server/localfile.go
```diff
@@ -132,89 +153,153 @@ func (s *server) Read(req *pb.ReadActionRequest, stream pb.LocalFile_ReadServer)
...
+		fileChan <- file
+		close(fileChan)
+	}
+
+	// TODO: Make it configurable.
+	parallelism := 3
+	for i := 0; i < parallelism; i++ {
+		errs.Go(func() error {
+
+			for {
```
sfc-gh-psmolenski, line 179: Extract the content of `for` to a separate function to improve readability

services/localfile/server/localfile.go
sfc-gh-mwalas: You could argue we get the List implementation done using `filepath.Glob` too, but since this codebase is already written there to be ready for recursive option and compatible with streaming of large number of files I wanted just to reuse List.

That being said I do not see clearly why list didn't got, at the fist cut, implemented with `filepath.Glob`, I do sense there might be some history there.

sfc-gh-mwalas: Possibly the consideration was that Glob eats errors:

`Glob ignores file system errors such as I/O errors reading directories. The only possible returned error is [ErrBadPattern](https://pkg.go.dev/path/filepath#ErrBadPattern), when pattern is malformed.`

services/localfile/server/localfile.go
sfc-gh-mwalas: I've tried doing it but the code didn't end up more readable in the end to be honest: the internals depend on many vars allocated at the Read-rpc level.

We could tackle this by having a separate RPC for the "grep" verb that internally calls out to a single-file read. But saying this I don't like performance implications of such move.

Hence, I propose to leave it as it is in terms of structure for the time being.

services/services.go
```diff
@@ -49,3 +56,31 @@ func ListServices() []SansShellRPCService {
 	defer mu.RUnlock()
 	return rpcServices
 }
+
+// SetAPIVersion sets currently supported API version.
+// Versioning must follow https://semver.org/.
+func SetAPIVersion(version string) error {
+	mu.RLock()
+	defer mu.RUnlock()
+	if _, ok := supportedAPIVersions[version]; !ok {
```
sfc-gh-ikryvanos, line 65: In function description you specify `supportedAPIVersions` includes `v1` and `v2` which is not semver. Semver should not start with `v` and have major, minor and patch.

Let's either remove semver requirements or change version format

sfc-gh-ikryvanos (changes requested):

README.md
```diff
@@ -225,6 +225,32 @@ time.
...
+be changed each time we add a backward-incompatible change to Sansshell services.
+
+Default version supported by Sanasshell server is set to `v1`, in order to use
+features of higher API version you should audit your OPA policy to check if there
+are no unintentional side-effects of allowing new Sansshell features.
+
+#### List of current API versions
+
+  - `v1`
+  - `v2` -- allow to read a directory while specifying a trailing wildcard, for
```
sfc-gh-psmolenski, line 251: Shouldn't we tie API version to sansshell version, where the major version of sansshell is at the same time the API version?

cmd/sansshell-server/main.go
```diff
@@ -84,6 +85,7 @@ var (
 
 	policyFlag    = flag.String("policy", defaultPolicy, "Local OPA policy governing access.  If empty, use builtin policy.")
 	policyFile    = flag.String("policy-file", "", "Path to a file with an OPA policy.  If empty, uses --policy.")
+	apiVersion    = flag.String("api-version", "v1", "Version of the Sansshell services API accepted by the server. Policy set in proxy and server must be verified before upgrading against API extensions to avoid unintentional side effects.")
```
sfc-gh-psmolenski, line 88: Shouldn't the accepted API version be defined in the policy file? The only way to update the accepted API version would then be to modify the rego file, which would be a clear indication that this change may have affected client permissions and should, as such, be reviewed by policy owners. Whereas modifying the server flag may go unnoticed.

cmd/sansshell-server/default-policy.rego
```diff
@@ -21,7 +21,6 @@ allow {
 
 allow {
 	input.type = "LocalFile.ReadActionRequest"
-	input.message.file.filename = "/etc/hosts"
```
sfc-gh-psmolenski, line 24: Why did you remove this line?

README.md
sfc-gh-psmolenski: I see your point. Maybe we could add a custom option to Protobuf message definition with API version for specific fields.

sfc-gh-psmolenski (approved):

services/services.go
sfc-gh-mwalas: Sure, I've put `v1` as this got parsed by the hashicorp semver lib. Better to stick to pure semver though.

cmd/sansshell-server/default-policy.rego
sfc-gh-mwalas: For a better local testing of the read feature over multiple files (arguably it is impossible to test it if we limit ourselves to only be able to read `/etc/hosts`).

cmd/sansshell-server/main.go
sfc-gh-psmolenski: Having a command line flag is a good thing, but being able to mark REGO policy as compatible with a given API version is equally handy. In the end, the goal is to make accidental API upgrades impossible or very unlikely, which means:
1. If you are deploying Sansshell for the first time, you should be using the latest API by default.
2. If you are upgrading Sansshell in an existing deployment, you should keep compatibility with the same API version you had before the upgrade, unless you specifically opt into upgrading API (either through flags or REGO)

sfc-gh-mwalas: If we state [1] as requirement then command line flag is clearly not a good fit for controlling those as the only way. Yet if we follow the plan I stated above we can add that behaviour fairly easily through another version in default REGO that would take precedence if it is higher than cmd line flag.

That sort of convinces me we can resolve this comment and record follow-ups to versioniong:
 - add version specification in REGO;
 - add a proto option to be used with new fields for version compatibility checking;

sfc-gh-ikryvanos (approved):

PR #536, opened 2025-02-25
Add get versions function

sfc-gh-ikryvanos:
Add function to get sansshell-server version of all targets from con

### DIFF
--- a/services/sansshell/client/utils.go
+++ b/services/sansshell/client/utils.go
@@ -48,8 +48,7 @@ func GetSansshellVersion(ctx context.Context, conn *proxy.Conn) (string, error) 
 
 // GetSansshellVersionForMany gets the version of the sansshell servers.
 // Returns error if:
-// - was provided not exactly one target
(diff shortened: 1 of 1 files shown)


services/sansshell/client/utils.go
```diff
@@ -45,3 +45,18 @@ func GetSansshellVersion(ctx context.Context, conn *proxy.Conn) (string, error)
 
 	return resp.GetVersion(), nil
 }
+
+// GetSansshellVersionForMany gets the version of the sansshell servers.
+// Returns error if:
+// - was provided not exactly one target
```
sfc-gh-psmolenski, line 51: Isn't the whole point of the function to accept more than one target ;)

sfc-gh-psmolenski (approved):

services/sansshell/client/utils.go
sfc-gh-ikryvanos: blindly copy-past, thanks

PR #543, opened 2025-03-13
Add authz policy dry run

sfc-gh-ikryvanos:
Adding ability to execute command in authz dry run mode


proxy/server/target.go
```diff
@@ -304,6 +307,13 @@ func (s *TargetStream) Run(nonce uint32, replyChan chan *pb.ProxyReply) {
 				return err
 			}
 
+			if s.authzDryRun {
+				// TODO: make authz dry run request to server
+				st := status.Newf(codes.Aborted, "authz dry run passed. Proxy: Ok, Server: Unknown. Aboarting \"%s\" futher execution", s.Method())
```
sfc-gh-mwalas, line 312: s/Aboarting/Aborting

cmd/sanssh/main.go
sfc-gh-mwalas, line 1: It seems that we return somewhat clear error message for access deny:

```
go run ./cmd/sanssh --authz-dry-run --proxy=localhost:50043 --targets=localhost:50042 file read /etc/adsf 

Target localhost:50042 (0) returned error - rpc error: code = PermissionDenied desc = OPA policy does not permit this request: this message always shows up on errors, we only proxy /etc/hosts
exit status 1
```

but in case of authz policy pass what we print is confusing:

```
go run ./cmd/sanssh --authz-dry-run --proxy=localhost:50043 --targets=localhost:50042 file read /etc/hosts

Target localhost:50042 (0) returned error - rpc error: code = Aborted desc = authz dry run passed. Proxy: Ok, Server: Unknown. Aboarting "/LocalFile.LocalFile/Read" futher execution
exit status 1
```


I would hope that for authz check pass case we could override the printed error, possibly at https://github.com/Snowflake-Labs/sansshell/blob/7f65cbc0bb10010741ad75400ed4b7e2d40a91d5/cmd/sanssh/client/client.go#L151

While Aborted is correct code to be passed from server to client, it is confusing for the reader a bit. Even printing some more human-friendly information about what happened there additionally to ```Target localhost:50042 (0) returned error - rpc error: code = Aborted desc = authz dry run passed. Proxy: Ok, Server: Unknown. Aboarting "/LocalFile.LocalFile/Read" futher execution``` might be enough.

sfc-gh-ikryvanos: I have just made some changes in client code. Now output for single target looks like this 
```sh
go run ./cmd/sanssh --proxy=localhost:50043 --targets=localhost:50042 --authz-dry-run file read /etc/hosts
localhost:50042: Authz dry run passed. Proxy: Ok, Server: Unknown. Aborting "/LocalFile.LocalFile/Read" futher execution
Target localhost:50042 (0) returned error - rpc error: code = Aborted desc = Aborted by proxy
exit status 1
```

I can't significantly impact on error message at the very end, because it is constructed by each command separately. Or it would require to update all clients commands handlers, what I want to avoid. 

For me it looks clear, probably because I made this message, from the side does it looks clear, too?

proxy/proxy.proto
```diff
@@ -84,6 +84,9 @@ message StartStream {
   // to that timeout before proceeding. Each Dial() happens in their
   // own routine so this will not block the overall progress for a stream.
   google.protobuf.Duration dial_timeout = 4;
+
+  // Flag to execute rpc in authz dry run mode
```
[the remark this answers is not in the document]
sfc-gh-ikryvanos: Updated

cmd/sanssh/main.go
sfc-gh-mwalas: Thanks, it is better. Seems like 2 lines per target would be emitted only for authz dry run which is good way of restricting this behavior.

sfc-gh-mwalas (approved):

PR #545, opened 2025-03-24
Make rpc authz configurable

sfc-gh-ikryvanos:
Get rid of hard coded dependency of OPA auth and make rpcauth mechanism configurable


auth/opa/opa.go
```diff
@@ -202,3 +203,17 @@ func (q *AuthzPolicy) DenialHints(ctx context.Context, input interface{}) ([]str
 	sort.Strings(hints)
 	return hints, nil
 }
+
+// NewOpaRPCAuthorizer creates a new Authorizer from a policy string. Any supplied
+// authorization hooks will be executed, in the order provided, on each policy
+// evaluation.
+// NOTE: The policy is used for both client and server hooks below. If you need
+//
```
sfc-gh-mwalas, line 211: nit: remove extra empty line in comment

auth/rpcauth/rpcauth.go
```diff
@@ -47,14 +48,43 @@ var (
...
+	// AuthorizeStream implements grpc.StreamServerInterceptor and applies policy checks on any RecvMsg calls to a service
+	AuthorizeStream(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error
+
+	// AuthorizeClient implements grpc.UnaryClientInterceptor, and will authorize each rpc on each call to remote service
+	AuthorizeClient(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error
+
+	// AuthorizeClientStream implements grpc.StreamClientInterceptor and applies policy checks on any SendMsg calls to remote service
+	AuthorizeClientStream(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error)
+
+	// AppendHooks adds additional hooks to the authorizer to execute before authz policy evaluation
```
sfc-gh-mwalas, line 76: Seems like that is the only new method added to the interface and it is unused in this PR, could you explain somewhere, maybe in the PR description how this one would be used?

auth/rpcauth/rpcauth.go
```diff
@@ -31,10 +31,11 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 
-	"github.com/Snowflake-Labs/sansshell/auth/opa"
 	"github.com/Snowflake-Labs/sansshell/telemetry/metrics"
 )
 
+//go:generate mockgen -source=./rpcauth.go -destination=./rpcauth_mock.go -package=rpcauth AuthzPolicy
```
sfc-gh-mwalas, line 37: This mock doesn't seem to be used anywhere, why do we need it now?

Also, if you mock out `RPCAuthorizer` the tests won't be that meaningful, are you thinking of mocking `AuthzPolicy`? In any case I feel like we gain much better test coverage by generating some interesting policy cases (actual OPA) and testing actual implementations on real policy.

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.28.0
 	go.opentelemetry.io/otel/sdk/metric v1.28.0
 	go.opentelemetry.io/otel/trace v1.28.0
+	go.uber.org/mock v0.4.0
```
sfc-gh-mwalas, line 27: That is the first time we add generated mocks to sansshell-labs, can we talk about it a little bit?

auth/opa/opa.go
```diff
@@ -202,3 +203,17 @@ func (q *AuthzPolicy) DenialHints(ctx context.Context, input interface{}) ([]str
...
+//	distinct policy for client vs server, create 2 Authorizer's.
+func NewOpaRPCAuthorizer(ctx context.Context, opaPolicy string, authzHooks ...rpcauth.RPCAuthzHook) (rpcauth.RPCAuthorizer, error) {
+	authzPolicy, err := NewAuthzPolicy(ctx, opaPolicy)
```
sfc-gh-mwalas, line 214: seems like opa just calls to rpcauth which has generic implementation (non-OPA) of Authorizer based on OPA-specific `AuthzPolicy`. It would be clearer if here we would not create `authzPolicy` through `NewAuthzPolicy` but rather have 

`opaAuthzPolicy, err := NewOpaAuthzPolicy(ctx, opaPolicy)`

Speaking of this, is this going to sustain the test of time? I don't see where it would fail clearly as of now,  but need another round to convince myself.

auth/opa/opa.go
```diff
@@ -154,7 +155,7 @@ func NewAuthzPolicy(ctx context.Context, policy string, opts ...Option) (*AuthzP
 // Eval evaluates this policy using the provided input, returning 'true'
 // iff the evaulation was successful, and the operation represented by
 // `input` is permitted by the policy.
-func (q *AuthzPolicy) Eval(ctx context.Context, input interface{}) (bool, error) {
+func (q *AuthzPolicy) Eval(ctx context.Context, input *rpcauth.RPCAuthInput) (bool, error) {
```
sfc-gh-mwalas, line 158: can't we hide implementation of opa's `AuthzPolicy` struct by naming it `opaAuthzPolicy`?

sfc-gh-mwalas (commented):
I think I roughly got what you are up to here, but need another round to get through details.

auth/rpcauth/rpcauth.go
sfc-gh-ikryvanos: I mock only `AuthPolicy` and used it in rpcauth_test.go. It is required to test `RPCAuthorizer` without OPA as dependency, otherwise it would be cycle dependency problem.

auth/opa/opa.go
sfc-gh-ikryvanos: renamed.

>Speaking of this, is this going to sustain the test of time? I don't see where it would fail clearly as of now, but need another round to convince myself.

it how it works before and until we won't change authz logic itself it would remain the same. Now both `AuthzPolicy` and `RPCAuthorizer` is abstract interfaces and any changes in internal logic won't affect other part of project

auth/opa/opa.go
sfc-gh-ikryvanos: removed

auth/rpcauth/rpcauth.go
sfc-gh-ikryvanos: Also I changed command to generate only `AuthPolicy` mock

auth/opa/opa.go
sfc-gh-ikryvanos: done

auth/rpcauth/rpcauth.go
sfc-gh-ikryvanos: Its an artifact left from last refactoring. Removing it

auth/opa/opa.go
sfc-gh-mwalas: Sure, thanks.

auth/rpcauth/rpcauth.go
sfc-gh-mwalas: Ah sorry, it didn't render large diff for me and I missed `rpcauth_test.go`. Will look now, thanks.

auth/rpcauth/rpcauth_test.go
```diff
@@ -400,11 +390,22 @@ func TestAuthzHook(t *testing.T) {
...
+
+			authzMockEval:   true,
+			authzEvalResult: false,
 		},
 	} {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
-			authz := New(policy, tc.hooks...)
+			ctrl := gomock.NewController(t)
+			authzPolicyMock := NewMockAuthzPolicy(ctrl)
```
sfc-gh-mwalas, line 401: I am always torn when I see in golang those mocks used and then mentally compare that to defining for test a simple struct which could have fields
```
authzMockEval bool
authzEvalResult bool
```

and be created as part of testcase like:

```
authzStub {
  authzMockEval:   true,
  authzEvalResult: false,
}
```

I've witnessed also in the past a bit crack-down on mock-ing interfaces and framework in favor of simple, inline stubs (or Fakes) created for tests. Saying this I am using definition from https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs

This is partially a style / preference issue, yet there are some notable differences: with mocks you test (and assert) behavioral contract strongly creating form of API contract; with stub it is more about whether the software "does what is supposed to", as stub can store some state and later assert on the internal state for entire case.

Typically you want small number of highly contractural interfaces (enforced with mocking) in somewhat monolithic codebase, as having many of those means that you have to redo your tests heavily when you change the contract. Those make sense though if you have some external implementations injected and want to verify that regardless of those the code you have at hand does the correct thing. For example, we could test interactions of the "sansshell service registration" using mock-service so that it is independent of actual service impl.

That all being said, I am interested in your reasoning for using mocks in this particular case vs a simple stub / fake.

cmd/proxy-server/main.go
```diff
@@ -125,7 +125,18 @@ func main() {
 	ctx := logr.NewContext(context.Background(), logger)
 	ctx = metrics.NewContextWithRecorder(ctx, recorder)
 
-	parsed, err := opa.NewAuthzPolicy(ctx, policy, opa.WithDenialHintsQuery("data.sansshell.authz.denial_hints"))
+	// Optional authz policy for outbount requests
+	var clientOpaAuthzPolicy rpcauth.AuthzPolicy
+	if clientPolicy != "" {
+		var err error
```
sfc-gh-mwalas, line 131: I think you can skip declaring `err` here. Why setting err defined at top-level main() in the next line to `nil` would be a bad thing?

go.mod
sfc-gh-mwalas: Discussed in another comment.

sfc-gh-mwalas (approved):
I was trying to vet if this doesn't introduce any authz-bypass bug and seems like it is correct, but would appreciate 3rd pair of eyes to look at this code this way too.

cmd/proxy-server/main.go
sfc-gh-ikryvanos: We could, let me remove it

auth/rpcauth/rpcauth_test.go
sfc-gh-ikryvanos: Actually I used here mock as stub. Just provide some canned results for the calls. `Mockgen` I used to be able to reuse same double across the project (if it would be necessary)

sfc-gh-mwalas: FWIW if we prefer `fake` then mocking interfaces most of the time do not produce simplest way of injecting those, especially in golang where we could simply have test-only impl of interfaces. I do prefer fake's too https://github.com/snowflakedb/access-group-manager/blob/main/lib/crypto/cryptotest/cryptotest.go

PR #559, opened 2025-04-23
Add DNS lookup latency to the response

sfc-gh-mszczepanski:
Adding the time it took to make a DNS lookup as a field in the response.


sfc-gh-mwalas (approved):
Thanks!

PR #560, opened 2025-04-28
Remove defunct FDB emergency movement commands

sfc-gh-abeamon:
This removes the already unusable FDB emergency movement commands from sansshell


sfc-gh-abeamon:
From the incompatible protoc changes lint:

```
Error: Previously present message "FDBCLITenantEmergencyMove" was deleted from file.
Error: Previously present message "FDBCLITenantEmergencyMoveAbort" was deleted from file.
Error: Previously present message "FDBCLITenantEmergencyMoveFinish" was deleted from file.
Error: Previously present message "FDBCLITenantEmergencyMoveStart" was deleted from file.
Error: Previously present message "FDBCLITenantEmergencyMoveStatus" was deleted from file.
Error: Previously present message "FDBCLITenantEmergencyMoveSwitch" was deleted from file.
Error: Previously present message "FDBMoveDataCopyRequest" was deleted from file.
Error: Previously present message "FDBMoveDataCopyResponse" was deleted from file.
Error: Previously present message "FDBMoveDataWaitRequest" was deleted from file.
Error: Previously present message "FDBMoveDataWaitResponse" was deleted from file.
Error: Previously present service "FDBMove" was deleted from file.
Error: Previously present field "49" with name "tenant_emergency_move" on message "FDBCLICommand" was deleted.
```

Per recommendation elsewhere, I removed these without deprecation due to the fact that they are currently unusable anyway. If there are alternate suggestions to bypass the check or make a different change, I can do so.

services/fdb/fdb.pb.go
```diff
@@ -552,7 +308,7 @@ type FDBCLIAdvanceversion struct {
 func (x *FDBCLIAdvanceversion) Reset() {
 	*x = FDBCLIAdvanceversion{}
 	if protoimpl.UnsafeEnabled {
-		mi := &file_fdb_proto_msgTypes[9]
+		mi := &file_fdb_proto_msgTypes[5]
```
sfc-gh-abeamon, line 311: I’m not super familar with how the protobuf generation works, is it problematic that we are changing the indexes of unrelated items?

services/fdb/fdb.proto
```diff
@@ -606,7 +539,6 @@ message FDBCLICommand {
     FDBCLISleep sleep = 22;
     FDBCLIStatus status = 23;
     FDBCLISuspend suspend = 43;
-    FDBCLITenantEmergencyMove tenant_emergency_move = 49;
```
sfc-gh-mwalas, line 609: Since we are removing a field could we make the number and name reserved https://protobuf.dev/programming-guides/proto3/#fieldreserved

```
  reserved 49;
  reserved "tenant_emergency_move";
```

services/fdb/fdb.pb.go
sfc-gh-mwalas: I don't think this should be dangerous, we reduced number of types defined in proto by removing obsolete messages so the slice became shorter / more compact.

One way we could mitigate it for the time being would be to keep old message definitions around with prefix `OBSOLETE_`

services/fdb/fdb.proto
sfc-gh-abeamon: I tried to mark it reserved, but that continued to fail the incompatible changes lint. For now I've marked all removed messages/services as deprecated instead and I left this particular field in place.

Is this acceptable, or should I make a different sort of change?

sfc-gh-mwalas: Thanks, we should be good here!

PR #563, opened 2025-05-13
Redacting `authorization` header from grpc metadata

sfc-gh-psmolenski:
With this change, we filter out `authorization` grpc header from logs.

In the future, we may introduce a more generic approach where a user can define which headers should be treated.


auth/rpcauth/redact.go
```diff
@@ -198,3 +212,11 @@ func getRedactedInput(input *RPCAuthInput) (RPCAuthInput, error) {
 	redactedInput.Message = json.RawMessage(marshaled)
 	return redactedInput, nil
 }
+
+func isRedactedMetadata(key string) bool {
+	if strings.ToLower(key) == "authorization" {
```
sfc-gh-mwalas, line 217: I do think golangci is right here, seems like `return strings.ToLower(key) == "authorization"` is much more elegant here.

sfc-gh-mwalas (approved):

PR #574, opened 2025-06-16
SNOW-2147639: fix fdb conf command when key string has a column symbol

sfc-gh-gnugmanov:
This PR is fixing the `fdb conf` command. The problem is that INI library is treating column `:` as a delimiter. In FDB config we have a key string which has a column and that causes big problems.

The solution is to initialize the library with only `=` symbol as a delimiter. Also we need to remove the backticks when writing the target key starting as FDB behavior  is undefined in this case.

### DIFF
--- a/services/fdb/server/conf_test.go
+++ b/services/fdb/server/conf_test.go
@@ -63,6 +63,17 @@ func TestRead(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:      "read key with colon in name",
+			respValue: "test_value",
+			req: &pb.ReadRequest{
+				Location: &pb.Location{
+					File:    path,
+					Section: "backup_agent",
+					Key:     "the:key:with:colon",
+				},
+			},
+		},
 	} {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


sfc-gh-gnugmanov:
integration test passes
```
    bazel-bin/cmd/sanssh/sanssh_/sanssh -t $TARGET -proxy $DEPLOYMENT --justification="SNOW-994784 PSAC-FDB_Maintenance" fdb conf write fdbserver key1 value1
    bazel-bin/cmd/sanssh/sanssh_/sanssh -t $TARGET -proxy $DEPLOYMENT --justification="SNOW-994784 PSAC-FDB_Maintenance" fdb conf write fdbserver keypart1:keypart2 value2
    bazel-bin/cmd/sanssh/sanssh_/sanssh -t $TARGET -proxy $DEPLOYMENT --justification="SNOW-994784 PSAC-FDB_Maintenance" file read /etc/foundationdb/foundationdb.conf | grep value
    echo === Deleting ===
    bazel-bin/cmd/sanssh/sanssh_/sanssh -t $TARGET -proxy $DEPLOYMENT --justification="SNOW-994784 PSAC-FDB_Maintenance" fdb conf delete fdbserver keypart1:keypart2
    bazel-bin/cmd/sanssh/sanssh_/sanssh -t $TARGET -proxy $DEPLOYMENT --justification="SNOW-994784 PSAC-FDB_Maintenance" fdb conf delete fdbserver key1
    bazel-bin/cmd/sanssh/sanssh_/sanssh -t $TARGET -proxy $DEPLOYMENT --justification="SNOW-994784 PSAC-FDB_Maintenance" fdb conf delete fdbcluster key1
    bazel-bin/cmd/sanssh/sanssh_/sanssh -t $TARGET -proxy $DEPLOYMENT --justification="SNOW-994784 PSAC-FDB_Maintenance" file read /etc/foundationdb/foundationdb.conf | grep value
=== Adding config ===
key1                                                           = value1
keypart1:keypart2                                              = value2
=== Deleting ===
```

sfc-gh-kkozielczyk:
Can you please update test input to include the case you are addressing, so we (1) know it's fixed, and (2) won't break it again in the future.

Bonus points for adding some more tricky configs to the input, like values with `/` or `\` etc. (whatever is supported by FDB config files)

sfc-gh-ldormoy:
Thanks Gaziz! The change LGTM. I agree with @sfc-gh-kkozielczyk's comment on testing.

services/fdb/server/conf_test.go
```diff
@@ -283,3 +294,393 @@ cluster_file = /etc/foundatindb/fdb.cluster`,
 		})
 	}
 }
+
+func TestWriteWithColonInKey(t *testing.T) {
```
sfc-gh-kkozielczyk, line 298: Can those test cases be merged into the existing `TestWrite` function? Is there anything requiring a separate function for those? Same for the other three new test functions, why not just add test cases to the existing ones? (sorry if there's some special test setup, didn't notice anything that can't be part of the existing test functions)

sfc-gh-gnugmanov:
pinged in slack

sfc-gh-ikryvanos (approved):

PR #578, opened 2025-07-23
Add flag for setting stream response for httpoverrpc proxy cmd

sfc-gh-mjankowski:
Added a flag in the same manner as for httpoverrpc `get` command.
Implemented tests for new not streamed scenario.


services/httpoverrpc/client/client.go
```diff
@@ -120,6 +121,53 @@ func validatePort(port int) bool {
...
+		if rs.GetBody() != nil {
+			if _, err := httpResp.Write(rs.GetBody()); err != nil {
+				fmt.Fprintln(os.Stdout, err)
+			}
+		}
+	}
+}
+
+func handleProxyHost(ctx context.Context, httpResp http.ResponseWriter, proxy pb.HTTPOverRPCClientProxy, req *pb.HostHTTPRequest) {
+	resp, err := proxy.Host(ctx, req)
```
sfc-gh-mwalas, line 155: if I read this correctly we used to call, prior to adding the flag `resp, err := proxy.StreamHost(ctx, req)` below (seeing this in removed line), now with flag set to default value "false" we will use instead `proxy.Host(ctx, req)`

Is it intended that this PR changes default behavior?

sfc-gh-mjankowski: It turned out that previous change (not adding a flag was) a mistake, because if for some reasons sansshell server is not updated to the recent version it might not contain the `StreamHost` method. So now in this situation it will be working by default and someone might use `--stream` flag to streamline the communication.

sfc-gh-mjankowski: For `Get` command the flag was added, it was only missing for the `Proxy`.

sfc-gh-mwalas (approved):

PR #588, opened 2025-09-03
Add URIs to input

sfc-gh-ikryvanos:
Add URIs to input auth info to be able to use it in authz policy

### DIFF
--- a/auth/rpcauth/input.go
+++ b/auth/rpcauth/input.go
@@ -146,7 +146,7 @@ type CertAuthInput struct {
 	// DNS names, from SubjectAlternativeName
 	DNSNames []string `json:"dnsnames"`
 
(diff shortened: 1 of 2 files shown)


auth/rpcauth/input.go
```diff
@@ -145,6 +146,9 @@ type CertAuthInput struct {
 	// DNS names, from SubjectAlternativeName
 	DNSNames []string `json:"dnsnames"`
 
+	// URIs from cetificate
```
sfc-gh-mwalas, line 149: similarly to the comment at `DNSNames` I would state here that these are  "`URIs` from `SAN URI` certificate extensions"

sfc-gh-mwalas (approved):

PR #590, opened 2025-09-08
Add WithServerOption function

sfc-gh-rmaj:
Adding a way to pass options directly to the `grpc.NewServer`. It will allow to set up server-instruments: https://grpc.io/docs/guides/opentelemetry-metrics/#server-instruments, as shown in the example: https://github.com/grpc/grpc-go/blob/master/examples/features/opentelemetry/server/main.go#L73-L93


sfc-gh-kyurtsever (commented):
I think you don't want this to be configurable at a per service level since it will affect the whole server. Therefore you should be able to just modify the place where the server is set up instead of adding a new function for that.

PR #613, opened 2026-03-16
Implement multi identity caller that looks at the server cert while selecting the client one

sfc-gh-mjankowski:
## The mTLS handshake and where we inject new logic

The diagram below shows the messages exchanged during a standard mTLS
handshake (TLS 1.2 flow shown for clarity; TLS 1.3 reorders some messages
but the certificate selection logic is equivalent). The **highlighted step**
is where `MultiIdentityCredentials` injects its identity-selection logic.

```
  Proxy (client)                                    Target (server)
  ─────────────                                     ───────────────
       │                                                  │
       │─── ClientHello ─────────────────────────────────▶│
       │    (supported ciphers, TLS version, random)      │
       │                                                  │
       │◀── ServerHello ──────────────────────────────────│
       │    (chosen cipher, TLS version, random)          │
       │                                                  │
       │◀── Server Certificate ───────────────────────────│
       │    (server's leaf cert + chain)                   │
       │                                                  │
       │    ┌─────────────────────────────────────────┐   │
       │◀───│ CertificateRequest                      │───│
       │    │  • AcceptableCAs: [CA-Acme, ...]        │   │
       │    │    (DER-encoded subject DNs of CAs the  │   │
       │    │     server trusts for client certs)      │   │
       │    └─────────────────────────────────────────┘   │
       │                                                  │
       │◀── ServerHelloDone ──────────────────────────────│
       │                                                  │
       │    ╔═════════════════════════════════════════╗   │
       │    ║  ** OUR NEW LOGIC RUNS HERE **          ║   │
       │    ║                                         ║   │
       │    ║  GetClientCertificate callback fires:   ║   │
       │    ║                                         ║   │
       │    ║  1. Read AcceptableCAs from the server  ║   │
       │    ║  2. For each identity we hold:          ║   │
       │    ║     - extract leaf cert's Issuer DN     ║   │
       │    ║     - check if it's in AcceptableCAs    ║   │
       │    ║  3. Return the first matching cert      ║   │
       │    ║     (or fall back to identity[0])       ║   │
       │    ╚═════════════════════════════════════════╝   │
       │                                                  │
       │─── Client Certificate ──────────────────────────▶│
       │    (the cert selected by our logic above)        │
       │                                                  │
       │─── ClientKeyExchange ───────────────────────────▶│
       │    (premaster secret encrypted with server key)  │
       │                                                  │
       │─── CertificateVerify ───────────────────────────▶│
       │    (signature proving we own the private key     │
       │     of the cert we just sent)                    │
       │                                                  │
       │─── [ChangeCipherSpec] ──────────────────────────▶│
       │─── Finished ────────────────────────────────────▶│
       │                                                  │
       │◀── [ChangeCipherSpec] ───────────────────────────│
       │◀── Finished ─────────────────────────────────────│
       │                                                  │
       │    ══════ Encrypted channel established ══════   │
       │                                                  │
```

In standard (single-identity) mTLS the client always sends the same
certificate regardless of what the server asks for. With
`MultiIdentityCredentials`, the `GetClientCertificate` callback inspects
the server's `CertificateRequest` and dynamically picks the right identity.

## Proxy server wiring

The proxy server's credential extraction was refactored to support two paths:

- **Single identity** (zero or one `clientCredSources`): preserves the existing `WrappedTransportCredentials` behavior with its own refresh/metrics/logging.
- **Multiple identities** (two or more `clientCredSources`): loads each source as a `ClientIdentity` and creates a `MultiIdentityCredentials`.

The server-side (incoming) credentials are unchanged and still use the single `credSource`.


auth/mtls/client.go
```diff
@@ -27,6 +27,13 @@ import (
 	"google.golang.org/grpc/credentials"
 )
 
+// ClientIdentity represents a single client identity (certificate + name)
+// that can be presented during a TLS handshake.
+type ClientIdentity struct {
+	Name string
```
sfc-gh-mwalas: If `Name` supposed to be `DistinguishedName` then let's call it this way.

But looking at the code it is more of a "Loader Name" right?

cmd/proxy-server/server/server.go
```diff
@@ -119,13 +120,29 @@ func WithTlsConfig(tlsConfig *tls.Config) Option {
 }
 
 // WithCredSource applies a registered credential source with the mtls package.
+// It is used for both the server-side (incoming) identity and, when no
+// WithClientCredSource options are provided, for the single outbound client
+// identity (via WrappedTransportCredentials). Mutually exclusive with
```
sfc-gh-mwalas: Wait, this is not 
```Mutually exclusive with
// WithClientCredSource for outbound credentials.
```

it is rather simply ignored on client leg, so there it is perfectly valid to have both `WithCredSource` and also `WithClientCredSource` as a config, in which case the former is on server side and latter on client, correct?

cmd/proxy-server/server/server.go
```diff
@@ -456,38 +473,57 @@ func Run(ctx context.Context, opts ...Option) {
...
+// When clientCredSources are set (WithClientCredSource), MultiIdentityCredentials
+// is used for dynamic identity selection based on the server's AcceptableCAs.
+// credSource is used as the rootCALoader for server certificate verification.
+//
+// Otherwise, when only credSource is set (WithCredSource), the single-identity
+// path is used via LoadClientCredentials / WrappedTransportCredentials.
 func extractClientTransportCredentialsFromRunState(ctx context.Context, rs *runState) (credentials.TransportCredentials, error) {
-	var creds credentials.TransportCredentials
-	var err error
+	hasMulti := len(rs.clientCredSources) > 0
```
sfc-gh-mwalas, line 488: I wonder if we have to bind so tightly the aspect of having "multi identity" on the client side with ability to customize at all the client identity, currently those are quite deeply bound, but I do wonder if we can make them separated out in this PR?

auth/mtls/client_multi_credentials.go
```diff
@@ -0,0 +1,348 @@
...
+	// MinVersion 1.2 is required by RFC 7540 §9.2.
+	// CipherSuites is left nil — Go's crypto/tls defaults are safe and
+	// grpc-go's applyDefaults only filters when CipherSuites == nil too.
+	tlsCfg := &tls.Config{
+		ServerName: serverName,
+		RootCAs:    rootCAs,
+		MinVersion: tls.VersionTLS12,
+		NextProtos: []string{"h2"},
+		GetClientCertificate: func(cri *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			idx := selectIdentity(cri, identities, m.logger)
```
sfc-gh-mwalas: Why `selectIdentity` would return an `idx` instead of just identity (and error if not found / matching potentially)?

auth/mtls/client_multi_credentials.go
```diff
@@ -0,0 +1,348 @@
...
+//
+// AcceptableCAs entries are DER-encoded CA subject DNs sent by the server
+// (per RFC 5246 §7.4.4 / RFC 8446 §4.3.2). We check the RawIssuer of every
+// certificate in the chain against this set — matching Go's built-in behavior
+// in crypto/tls when selecting from tls.Config.Certificates.
+//
+// Falls back to index 0 if the server sends an empty list or no match is found.
+func selectIdentity(cri *tls.CertificateRequestInfo, identities []ClientIdentity, logger logr.Logger) int {
+	if len(cri.AcceptableCAs) == 0 || len(identities) <= 1 {
+		return 0
```
sfc-gh-mwalas: this will panic in the caller if identites is truly of length 0

auth/mtls/client_multi_credentials.go
```diff
@@ -0,0 +1,348 @@
...
+	logger.Info("no identity matched server's AcceptableCAs, falling back to first identity",
+		"identities", identityNames,
+	)
+	return 0
+}
+
+// chainMatchesAcceptableCAs checks whether any certificate in the identity's
+// chain has a RawIssuer that matches one of the server's AcceptableCAs.
+//
+// This mirrors Go's crypto/tls matching logic (tls.Config.Certificates):
```
sfc-gh-mwalas: If there are parts of this code that are, for whatever reason, forked from golang core implementation please add links to sources.

auth/mtls/client_multi_credentials.go
```diff
@@ -0,0 +1,348 @@
...
+// ClientHandshake performs the client-side TLS handshake. It uses
+// GetClientCertificate to pick the right client certificate based on the
+// server's CertificateRequest.AcceptableCAs. Server certificate verification
+// uses the rootCAs pool loaded from rootCALoader.
+//
+// Unlike WrappedTransportCredentials.ClientHandshake, refresh errors are
+// logged and the handshake proceeds with existing certs. In multi-identity
+// proxy scenarios availability is more important than cert freshness — a
+// transient loader failure should not block all outbound connections.
+func (m *MultiIdentityCredentials) ClientHandshake(ctx context.Context, serverName string, rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
```
sfc-gh-ssapkowski, line 197: This reimplementation of grpc-go internals is unnecessary. The only new feature here is `GetClientCertificate`, with CA traversal logic. 

You should build a new `tls.Config` with your `GetClientCertificate` and use https://pkg.go.dev/google.golang.org/grpc/credentials#NewTLS to build `TransportCredentials` and use its `ClientHandshake`.

```
type MultiIdentityCredentials struct {
    ...
    creds credentials.TransportCredentials
}

// buildCreds constructs inner creds from identities
// called at construction and on refresh
func (m *MultiIdentityCredentials) buildCreds(...) credentials.TransportCredentials {
    tlsCfg := &tls.Config{
       ...
       GetClientCertificate: func(...) { // here you plugin your selectIdentity }
    }
}

func (m *MultiIdentityCredentials) ClientHandshake(...) ... {
    ...
    conn, info, err := creds.ClientHandshake(ctx, serverName, rawConn)
}
```


All other methods like Info, Clone can also delegate to `creds`.

sfc-gh-ssapkowski (changes requested):

auth/mtls/client_multi_credentials.go
```diff
@@ -0,0 +1,348 @@
...
+// selectIdentity picks the first identity whose certificate chain is signed
+// by a CA the server trusts, by checking every certificate's RawIssuer against
+// the server's AcceptableCAs. This handles certificates issued by intermediate
+// CAs, where the leaf's immediate issuer DN won't appear in AcceptableCAs but
+// a higher-up CA in the chain will (e.g. the intermediate's issuer = the root).
+//
+// AcceptableCAs entries are DER-encoded CA subject DNs sent by the server
+// (per RFC 5246 §7.4.4 / RFC 8446 §4.3.2). We check the RawIssuer of every
+// certificate in the chain against this set — matching Go's built-in behavior
+// in crypto/tls when selecting from tls.Config.Certificates.
```
sfc-gh-ssapkowski: you claim that you match Go's built-in behavior from crypto/tls, but that's not exactly true - you are only comparing `RawIssuer` while `cri.SupportsCertificate` checks signature algorithm compability and issuer matching. I'd suggest delegating to `cri.SupportsCertificate` and fallback to raw issuer matching for intermediate CA scenarios

auth/mtls/client.go
sfc-gh-mjankowski: You're right, it's the credential loader's registration name, not a certificate DN. Renamed to LoaderName across the board.

cmd/proxy-server/server/server.go
sfc-gh-mjankowski: Good catch - the doc was wrong. WithCredSource and WithClientCredSource are not mutually exclusive.
Updated the comment.

auth/mtls/client_multi_credentials.go
sfc-gh-mjankowski: Agreed. Refactored selectIdentity to return (*ClientIdentity, error).

auth/mtls/client_multi_credentials.go
sfc-gh-mjankowski: Refactored, thx

auth/mtls/client.go
sfc-gh-ssapkowski: there is also implicit contract that 

> `ClientIdentity.Name` == a key in the global loader registry

I don't know if that's desired. Right now, one can do:
```
id := mtls.ClientIdentity{
    Name: "my-client-id",
    Cert: someCert,
}
m, _ := mtls.NewMultiIdentityCredentials([]ClientIdentity{id}, "root-loader", logger, nil)
```

and refresh will fail, because there is no such loader

sfc-gh-ssapkowski: this happens in tests:
```
"msg"="cert refresh failed, using existing certs" "error"="loader \"acme\": unknown loader type acme"
```

auth/mtls/client_multi_credentials.go
sfc-gh-mjankowski: Done, thx for suggestion.

auth/mtls/client_multi_credentials.go
```diff
@@ -0,0 +1,348 @@
...
+// dynamically selecting a client certificate during the TLS handshake based
+// on the server's CertificateRequest.AcceptableCAs. This allows a single
+// gRPC dial to present the correct identity without retry or fallback.
+//
+// The rootCAs pool used for server verification is loaded from a named
+// CredentialsLoader (rootCALoader) and refreshed together with the client
+// identities whenever any loader signals new certificates.
+type MultiIdentityCredentials struct {
+	mu           sync.RWMutex
+	identities   []ClientIdentity // GUARDED_BY(mu)
```
sfc-gh-ssapkowski: i smell cpp

auth/mtls/client.go
sfc-gh-mjankowski: Fixed, I am creating loaders in the constructor and returning error if the fails.

auth/mtls/client_multi_credentials.go
sfc-gh-mjankowski: Done, I will test whether it works in the same way (there are no edge cases on staging).

auth/mtls/client_multi_credentials.go
sfc-gh-mjankowski: Removed :D

auth/mtls/client_multi_credentials.go
sfc-gh-mjankowski: Done, I am first checking the `cri.SupportsCertificate` and later falling back to raw issuer matching.

auth/mtls/client_multi_credentials_test.go
sfc-gh-ssapkowski, line 1: The old version had `TestClientHandshake_WrongCertFails` — a test where the server trusts only "Other Corp" but the client has only an "Acme" identity, verifying the handshake is rejected. This is a valuable edge case (ensures the fallback-to-first-identity behavior doesn't silently succeed with the wrong cert). It's absent in the new version.

cmd/proxy-server/server/server.go
sfc-gh-mjankowski: I will handle this request in new PR.

auth/mtls/client_multi_credentials_test.go
sfc-gh-mjankowski: Done, I've tried to clean up tests a little bit and I've lost this one, thanks for spotting that.

sfc-gh-ssapkowski (approved):
👍

PR #618, opened 2026-04-10
Fix proxy hang when target is unreachable during client-streaming RPCs

sfc-gh-ikryvanos:
## What changed (user-visible)

- **Unreachable target:** `sanssh file cp` through the proxy to a target that doesn't exist (or is unreachable) now fails with a clear error instead of hanging indefinitely.
- **Target dies mid-upload:** if the target server crashes or is killed during a file upload, the client now reports an error and exits with a non-zero code instead of silently reporting 100% success.
- **Progress bar on error:** the progress bar is cleared when an upload fails, so users no longer see a confusing "100%" line followed by an error message.

## Why it was broken

`TargetStream.Send()` — the proxy method that enqueues data for forwarding to the target — only checked the gRPC stream's context for cancellation. Two problems:

1. **Before connection:** the gRPC stream context belonged to a placeholder (`unconnectedClientStream`) and was never cancelled when `NewStream` failed. `Send()` blocked forever on the full `reqChan`.

2. **After connection:** when the target died mid-stream, `grpcStream.SendMsg` returned `io.EOF` but the send loop exited without cancelling anything. `Send()` kept accepting data into the dead `reqChan` because the buffer had room (consumer gone) and the context wasn't cancelled yet.

## What was done

**`proxy/server/target.go`:**

- `Send()` now selects on both `s.ctx` (our own context, cancelled by `cancelFunc`) and `getStream().Context()` (gRPC transport-level cancellation). Both are needed because after `setStream()` they are different objects.
- `Run()` calls `s.cancelFunc()` when `NewStream` fails (was missing).
- `Run()` calls `s.cancelFunc()` when `SendMsg` returns `io.EOF` (target died mid-stream), so `Send()` fails immediately on the next call instead of silently enqueueing data.

**`services/localfile/client/client.go`:**

- Progress bar is cleared on error paths so it doesn't render 100% after a failure.

**`proxy/server/target_test.go`:**

- `TestSendUnblocksWhenTargetUnreachable` — verifies `Send` unblocks when the target is unreachable.
- `TestSendFailsAfterTargetDiesMidStream` — verifies `Send` returns an error after the target dies mid-stream.

### DIFF
--- a/proxy/server/target.go
+++ b/proxy/server/target.go
@@ -181,8 +181,16 @@ func (s *TargetStream) Send(req proto.Message) error {
 	select {
 	case s.reqChan <- req:
 		return nil
+	// s.ctx is cancelled by our cancelFunc — e.g. when the target dies
+	// mid-stream (SendMsg returns EOF) or when NewStream fails. Without
+	// this, Send would keep enqueueing data into a dead reqChan.
 	case <-s.ctx.Done():
 		return s.ctx.Err()
+	// getStream().Context() is cancelled by gRPC on transport-level
+	// failures. After setStream() it is a different object from s.ctx,
+	// so we need both to react to whichever fires first.
+	case <-s.getStream().Context().Done():
+		return s.getStream().Context().Err()
 	}
 }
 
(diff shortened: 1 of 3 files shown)


sfc-gh-jdubiel (approved):
LGTM

PR #621, opened 2026-06-30
Support source port binding in tcp-check

sfc-gh-mmusa:
## Context

- Adds a new optional flag in `network tcp-check` command to bind to a specific source port
- If not provided, it defaults to an ephemeral port
- Validates source port using the existing validator method

### Manual Validation

1. Run server `go run cmd/sansshell-server/main.go`.
2. Run a tcp listener (I used a basic python script).
3. Run command `go run ./cmd/sanssh --targets=localhost network tcp-check --source-port 10
000 localhost:3777`

Validated both success and failure (port already in use) paths.

### Integration Tests added

Added integration tests for both success and failure (port already in use) paths.


services/network/client/application/tcp-check.usecase.go
```diff
@@ -33,11 +33,13 @@ type tcpCheckUseCase struct {
 // - hostname is host of remote machine from which was checked tcp connectivity
 // - port is port of remote machine from which was checked tcp connectivity
 // - timeout is  seconds to wait for a response on remote machine from hostname
-func (p *tcpCheckUseCase) Run(ctx context.Context, hostname string, port uint32, timeoutSeconds uint) (<-chan *pb.TCPCheckManyResponse, error) {
+// - sourcePort is the local source port to bind (nil = OS-chosen ephemeral)
+func (p *tcpCheckUseCase) Run(ctx context.Context, hostname string, port uint32, timeoutSeconds uint, sourcePort *uint32) (<-chan *pb.TCPCheckManyResponse, error) {
```
sfc-gh-rszuma: What is the difference in this semantics between nil and 0 passed? Couldn't we have only one value for the port chosen port?

services/network/client/infrastructure/input/tcp-check.controller.go
```diff
@@ -63,12 +65,27 @@ func (p *TCPCheckCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...inte
...
 
+	var sourcePort *uint32
+	// Check if --source-port flag is set and if so, parse the value into a uint32
+	f.Visit(func(fl *flag.Flag) {
+		if fl.Name == "source-port" {
+			v := uint32(p.sourcePort)
+			sourcePort = &v
+		}
+	})
+	if sourcePort != nil {
```
sfc-gh-rszuma: In what scenario sourcePort will be nil if the default value is 0 and user can't provide nil value?

services/network/client/infrastructure/input/tcp-check.controller.go
```diff
@@ -63,12 +65,27 @@ func (p *TCPCheckCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...inte
...
 	}
 
+	var sourcePort *uint32
+	// Check if --source-port flag is set and if so, parse the value into a uint32
+	f.Visit(func(fl *flag.Flag) {
+		if fl.Name == "source-port" {
+			v := uint32(p.sourcePort)
+			sourcePort = &v
+		}
+	})
```
sfc-gh-rszuma: Why do we need that? Couldn't we change the SourcePort type in TCPCheckCmd to uint32 and not need this code or limited only to `sourcePort = &p.sourcePort` ?

services/network/client/application/tcp-check.usecase.go
sfc-gh-mmusa: It distinguishes two cases that would otherwise both map to `0`:
- `nil` → caller did not specify a source port
- `0` → caller explicitly requested port 0, which is a valid request but not supported

Collapsing both to `0` would also introduce unnecessary overhead. Previously, we relied on a single `connect` syscall, allowing the kernel to choose the port. This change essentially keeps that behavior for the unspecified case. Otherwise, using `0` as the default would require an additional `bind` syscall every time.

---

I hope that it also answers your following comments on implementation. Happy to discuss more if needed.

sfc-gh-rszuma: If it is valid request but not supported, shouldn't we change `func (p *TCPCheckCmd) SetFlags(f *flag.FlagSet) {` ?

services/network/client/infrastructure/input/tcp-check.controller.go
```diff
@@ -30,22 +30,24 @@ import (
...
 func (*TCPCheckCmd) Usage() string {
-	return `tcp-check <host>:<port>
+	return `tcp-check <host>:<port> [--source-port <port>] [--timeout <seconds>]
     Makes tcp connectivity check from the remote machine to specified server.
 `
 }
 
 func (p *TCPCheckCmd) SetFlags(f *flag.FlagSet) {
 	f.UintVar(&p.timeout, "timeout", 3, "Timeout in seconds to wait for response from --host on remote machine")
+	f.UintVar(&p.sourcePort, "source-port", 0, "Local source port to bind for the outgoing connection (1-65535). If unset, the OS chooses one.")
```
sfc-gh-rszuma: Isn't default value incorrect in comparison with description?

services/network/client/application/tcp-check.usecase.go
sfc-gh-mmusa: Unfortunately the flag module doesn't allow using `nil` as default. But as per our offline discussion, I have changed the implementation to use `0` for default value. PR updated.

services/network/server/infrastructure/output/tcp.client.go
```diff
@@ -33,15 +35,23 @@ type TCPClient struct {
...
 
 // CheckConnectivity is used to check tcp connectivity from remote machine to specified server
-func (p *TCPClient) CheckConnectivity(ctx context.Context, hostname string, port uint32, timeout time.Duration) (*app.TCPConnectivityCheckResult, error) {
+func (p *TCPClient) CheckConnectivity(ctx context.Context, hostname string, port uint32, timeout time.Duration, sourcePort uint32) (*app.TCPConnectivityCheckResult, error) {
 	hostToCheck := net.JoinHostPort(hostname, strconv.Itoa(int(port)))
 
 	dialer := net.Dialer{Timeout: timeout}
+	if sourcePort != 0 {
+		// Bind the outgoing connection to the requested local source port.
+		dialer.LocalAddr = &net.TCPAddr{Port: int(sourcePort)}
```
sfc-gh-dkrause, line 44: To confirm: the dialer is happy with getting only the port and no IP here?

sfc-gh-mmusa: Yes, confirmed it with both manual validation and integration tests

sfc-gh-ikryvanos, line 44: TIME_WAIT reuse gotcha. Fixed source port + no SO_REUSEADDR → after one OK check, socket sit in TIME_WAIT ~60s. Repeat check same --source-port fail EADDRINUSE ("Source port already in use") even though nothing listening. Surprise users scripting repeated checks. Fix: set SO_REUSEADDR via dialer Control fn, or doc the behavior.

services/network/client/infrastructure/input/tcp-check.controller.go
```diff
@@ -63,12 +65,20 @@ func (p *TCPCheckCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...inte
 		return subcommands.ExitUsageError
 	}
 
+	sourcePort := uint32(p.sourcePort)
```
sfc-gh-ikryvanos: Silent uint32 truncation bug. p.sourcePort is uint (64-bit). Cast happen BEFORE validation, so validator can't catch overflow. --source-port 4294967296 → 0 → silently "OS chooses". --source-port 4294967297 → 1 → bind wrong port. Fix: validate raw p.sourcePort against 1-65535 before cast.

sfc-gh-ikryvanos (changes requested):

services/network/server/infrastructure/output/tcp.client.go
sfc-gh-mmusa: Thanks for the callout. I have documented this behavior.

sfc-gh-ikryvanos (approved):

PR #622, opened 2026-07-02
Set SO_REUSEADDR for source port binding

sfc-gh-mmusa:
## Context

This PR follows-up on #621. Sets `SO_REUSEADDR` on the dialer socket via `net.Dialer.Control` so that a source port lingering in `TIME_WAIT` from a previous check can be reused immediately.

## Testing

Manually validated that the same source port can be reused in back-to-back repeated checks


sfc-gh-rszuma (approved):
LGTM

sfc-gh-ikryvanos (approved):

